PR #53, opened 2015-01-24
Fixing handling of relative library paths

MrCustomizer:
I would like to propose an improvement to the handling of lib_dir. I'm using platformio for an Arduino project and I try to keep the original directory structure as intended by the Arduino IDE. In that way, nobody is forced to use platformio, because Arduino IDE can be used to compile the project, too. I've already used the trick, described in https://github.com/ivankravets/platformio/issues/41 and renamed my main ino file to src.ino. It's a hack, but it works.

Now, there is only one thing left: The library path for the project has to be modified. It is one directory up in the path hierarchy. As the project is checked in to a version control system and platformio should work out of the box without adjusting any properties, I need to be able to set a relative directory path, and not an absolute one in my platformio.ini:

[platformio]
lib_dir = ../libraries

This doesn't work in the current version, because the path is used like this in the build directory, which is nested at least one level deeper in the directory structure. Therefore I propose this fix, which expands the path to an absolute one after reading it from the configuration file. I've already successfully tested this patch in Linux and Windows.

### DIFF
--- a/platformio/util.py
+++ b/platformio/util.py
@@ -4,7 +4,7 @@
 import json
 from os import name as os_name
 from os import getcwd, getenv, listdir, makedirs, utime
-from os.path import dirname, expanduser, isdir, isfile, join, realpath
+from os.path import dirname, expanduser, isdir, isfile, join, realpath, abspath
 from platform import system, uname
 from subprocess import PIPE, Popen
 
@@ -51,7 +51,7 @@ def get_lib_dir():
         config = get_project_config()
         if (config.has_section("platformio") and
                 config.has_option("platformio", "lib_dir")):
-            lib_dir = config.get("platformio", "lib_dir")
+            lib_dir = abspath(config.get("platformio", "lib_dir"))
             if lib_dir.startswith("~"):
                 return expanduser(lib_dir)
             else:
--- a/platformio/util.py
+++ b/platformio/util.py
@@ -51,11 +51,10 @@ def get_lib_dir():
         config = get_project_config()
         if (config.has_section("platformio") and
                 config.has_option("platformio", "lib_dir")):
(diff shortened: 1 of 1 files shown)


platformio/util.py
```diff
@@ -4,7 +4,7 @@
 import json
 from os import name as os_name
 from os import getcwd, getenv, listdir, makedirs, utime
-from os.path import dirname, expanduser, isdir, isfile, join, realpath
+from os.path import dirname, expanduser, isdir, isfile, join, realpath, abspath
```
ivankravets: PlatformIO uses [isort](https://github.com/timothycrosley/isort) for python imports. Please move `abspath` before `dirname` (alphabetic order)

platformio/util.py
```diff
@@ -51,7 +51,7 @@ def get_lib_dir():
         config = get_project_config()
         if (config.has_section("platformio") and
                 config.has_option("platformio", "lib_dir")):
-            lib_dir = config.get("platformio", "lib_dir")
+            lib_dir = abspath(config.get("platformio", "lib_dir"))
```
ivankravets: Need to rewrite this block of code. `abspath` doesn't work with `~` home signs. In this situation, `lib_dir` will be incorrect for the value `lib_dir = ~/Documents/Arduino/Libraries`.
1. Need to expand user's path
2. Then we can convert relative path to absolute. 

Can you change it?

P.S: Thanks a lot for your contribution! :+1:

MrCustomizer:
I will do the changes and come back here, later. ;)

MrCustomizer:
I've done the changes. Please double check, I'm not a Python professional ;).

ivankravets:
Thanks a lot! I've just merged your changes. 👍

PR #54, opened 2015-01-24
Add build path to ignore list

MrCustomizer:
I think it would be good to add the build path to the ignore list. Do you agree?


MrCustomizer:
Ok, thanks for the tip. I will take a look into this. I think, you can close this pull request without merging, then.

PR #205, opened 2015-05-17
Update qtcreator.rst

Lutino:
Edited the file to include an additional step for the initalization of the platformio project


ivankravets:
Thanks! 👍🏾

PR #235, opened 2015-06-13
Updated recommened platforms install command

ctag:
`$ platformio install [platform]`
The above line is obsolete, I've changed the recommendation to reflect this latest version of the program:
`$ platformio platforms install [platform]`


ivankravets:
Thanks! :+1:

PR #265, opened 2015-08-07
improved error wording

WillemMali:
Fixed a grammatical mistake, made the meaning  of the error clearer by replacing "it" with "this" and by adding an example.

It does increase the length of the output, but I think it's much easier to read and more informative.


ivankravets:
Thanks!

PR #333, opened 2015-11-22
Fix ESP-12E flash size

celsworth:
I believe this 512k maximum_size is in error; the 4M ldscript is used just above, and I had to manually bodge esptool's `-bz` parameter to play with ArduinoOTA.

This should set it to 4M correctly.


celsworth:
Er sorry I've no idea how all those branches got merged in..

ivankravets:
> I had to manually bodge esptool's -bz parameter to play with ArduinoOTA.

What do you mean under it? You can use own [extra_script](http://docs.platformio.org/en/latest/projectconf.html#extra-script).

celsworth:
I'm not sure what you mean. The -bz parameter to esptool sets the size of the flash, right? It is set to 512k on an ldscript that is configured for 4M, I thought they were supposed to match.

celsworth:
Thanks for merging this. However it seems to have produced a new issue:

```
/home/chris/.platformio/packages/tool-esptool/esptool -eo /home/chris/.platformio/packages/framework-arduinoespressif/bootloaders/eboot/eboot.elf -bo .pioenvs/esp12/firmware.bin -bm dio -bf 80 -bz 4096K -bs .text -bp 4096 -ec -eo .pioenvs/esp12/firmware.elf -bs .irom0.text -bs .text -bs .data -bs .rodata -bc -ec
error: invalid flash size value: 4096K
scons: *** [.pioenvs/esp12/firmware.bin] Error 255
```

It needs to be `4M` rather than `4096K`; this works:

```
/home/chris/.platformio/packages/tool-esptool/esptool -eo /home/chris/.platformio/packages/framework-arduinoespressif/bootloaders/eboot/eboot.elf -bo .pioenvs/esp12/firmware.bin -bm dio -bf 80 -bz 4M -bs .text -bp 4096 -ec -eo .pioenvs/esp12/firmware.elf -bs .irom0.text -bs .text -bs .data -bs .rodata -bc -ec
```
esptool seems to need the `M` suffix for anything bigger than 1M. I don't know where this should be fixed in the source though.

ivankravets:
Fixed in development version http://docs.platformio.org/en/latest/installation.html#development-version

Please re-test it. Thanks.

celsworth:
Thanks, all working now!

PR #374, opened 2015-12-13
OTA for Sublime Text

eliabieri:
I added a OTA upload method for sublime text. I have no idea how to
specify the IP in sublime text,  though.


ivankravets:
@eliabieri thanks a lot for PR! You don't need to create own command for it. Use http://docs.platformio.org/en/latest/projectconf.html#upload-port
```ini
[env:***]
....
upload_port = 192.168.0.1
```

PR #411, opened 2015-12-29
Allow requirements.txt to upgrade to minor versions.

skorokithakis:


ivankravets:
Thanks :) I've just moved it to setuptools https://github.com/platformio/platformio/commit/621393138f119c96a81bb0d1bdd4390ca421aab7

PR #418, opened 2015-12-30
updated URL in example to new mbed domain

dh1tw:
The URL for the file used in the mbed HTTP Client example has changed from http://mbed.org/media/uploads/donatien/hello.txt to https://developer.mbed.org/media/uploads/donatien/hello.txt


ivankravets:
Thanks, @dh1tw!

PR #420, opened 2015-12-31
Update platformio.sublime-project.tpl

sticilface:
Include SPIFFS upload in sublime project generation


ivankravets:
Thanks! 

P.S: @gepd is working on special Sublime Text package with GUI https://github.com/gepd/Deviot

Issue #427, opened 2016-01-02
Add generic targets to IDEs


ivankravets:
Each IDE should contain default targets/commands:

* Run/Build
* Clean
* Upload
* Upload using Programmer
* Upload SPIFFS image
* Update

### IDE

- [x] Eclipse
- [ ] CLion
- [ ] Sublime Text

PR #428, opened 2016-01-03
Add external programmer and SPIFFS targets to Emacs

ZachMassia:
As per #427.

Added in the remaining targets to my `CMakeLists.txt` file and updated the docs with the new key bindings.

I've already pushed the corresponding changes to the Emacs package: ZachMassia/@69d584003332908f49f25fb614ed67925a23a578


ivankravets:
Thanks! :+1: :blush:

PR #433, opened 2016-01-04
clang_complete support for Emacs and Atom

ZachMassia:
I've removed CMake from the Emacs template, and will be switching [PlatformIO-Mode](github.com/ZachMassia/PlatformIO-Mode) over to direct interaction with `platformio` once this is merged.

Also added is the initial Atom project template, which is just the `.gitignore` and `.clang_complete` for now. We will have to discuss further the direction to go with it.
Like I said on Gitter, we could implement a "Task Runner" (see https://atombuild.github.io/), but I'm not sure there is a need for both it and the existing plugin.


ivankravets:
Thanks!

PR #457, opened 2016-01-13
Fix #444

### LINKED ISSUE #444: Not clear how to overwrite $BOARD.options.resetmethod ..
I would like to force my board.options.resetmethod="ck", but see no way to do that outside of editing the board.json file .. is there a way to override this for my board in a project-local config?  (i.e. through platformio_extra_script.py - tried to override it like "esp12.board.options.resetmethod=ck" but no go .. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29712788-not-clear-how-to-overwrite-board-options-resetmethod?utm_campaign=plugin&utm_content=tracker%2F2014790&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F2014790&utm_medium=issues&utm_source=github).
</bountysource-plugin>

crushedice2000:
What a beautiful number!

### DIFF
--- a/docs/platforms/espressif_extra.rst
+++ b/docs/platforms/espressif_extra.rst
@@ -56,9 +56,9 @@ To override default LD script please use :ref:`projectconf_build_flags` from
 Custom Reset Method
 -------------------
 
-To overwrite predefined reset method for uploading please use :ref:`projectconf_extra_script`.
+To change the default reset method for uploading please use :ref:`projectconf`.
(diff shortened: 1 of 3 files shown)


ivankravets:
Thanks for the pull request. I have updated the code with the new option ``upload_resetmethod`` and etc. See https://github.com/platformio/platformio/commit/52ceb6475893b7d7764d05b2d7b89360fb639f74

Issue #480, opened 2016-01-28
Adding support for Lattice FPGAs and Icestorm opensource fpga toolchain


Obijuan:
In mid 2015, Clifford Wolf released the first opensource toolschain for FPGAs: [The icestorm project](http://www.clifford.at/icestorm/)

It would be great to include it into Platformio. Having an IoT ecosystem that include microprocessors, FPGAs and soft-processors will be amazing.  Platformio can be the first opensource tools to support that whole ecosystem.

We have started the integration. The results are in this github repo/wiki:

https://github.com/bqlabs/Platformio-FPGA/wiki/Platformio-FPGA-wiki-home

We initially have started with the support for the [Lattice icestick boards](http://www.latticesemi.com/icestick). We already have a working "hello world" FPGA example. When the command "platformio run" is executed, the leds.v verilog file is synthetized and the bitstream generated. With "platformio run --target uploaded" it is uploaded into the FPGA.

We need some tips on how is best way to proceed for integration into platformio. We will also need some tips on how to package the icestorm tools so that they can be automatically installed when the fpga project are created. We will do all the work, we just need some advices and documentation :-)

Thanks a lot

Obijuan:
We already have the package  toochain-icestorm-linux_x86_64-1.tar.gz, with all the icestorm tools compiled  for Linux_x86_64. How should be proceed? Where should we upload the package?
Thanks a lot

ivankravets:
@Obijuan Hi! Is difficult to provide toolchain for other host OS?

> Where should we upload the package?

With PlatformIO 3.0 you can keep it in own location. I'm working on it. However, if you don't wait for PlatformIO 3.0 I can place it to current [PlatformIO Storage](https://sourceforge.net/projects/platformio-storage/). Where I can download it?

Obijuan:
Thanks @ivankravets!

> Hi! Is difficult to provide toolchain for other host OS?

No. We will do it. But first we want to develop under linux, to understand all the details of platformio. Once everything is working fine, we will provide packages for the other host OS

> With PlatformIO 3.0 you can keep it in own location. I'm working on it. However, if you don't wait for PlatformIO 3.0 I can place it to current PlatformIO Storage.  

That would be great :-)

> Where I can download it?

[toochain-icestorm-linux_x86_64-1.tar.gz](https://github.com/bqlabs/Platformio-FPGA/raw/master/packages/toochain-icestorm-linux_x86_64-1.tar.gz)

Thanks a lot :-)

[...]

PR #482, opened 2016-01-28
Initial support for the Lattice FPGAs

Obijuan:
Initial work for the issue #480 
It is still in development, but the hello world verilog example is working ok in the icestick board

The icestorm tools should be installed externally.  We have an initial .tar.gz file for Linux, but we need tips on how to manage the packages


ivankravets:
@Obijuan Thanks a lot for you efforts!!! 

Please take a look at https://github.com/platformio/platformio/issues/479

I hope that I release PlatformIO 2.8.0 today. It will be the last major release in 2.x. I'm going to start work on PlatformIO 3.0 and maybe the first issue will be https://github.com/platformio/platformio/issues/479

In this case I'll try with you to create external platform with all required info. I'll answer you.

Obijuan:
Thanks for the information :-) I'll wait for your answer

Issue #456, opened 2016-01-13
Board compatibility with more than one platform


WisniewskiP:
At this moment it is only possible to configure board with one compatible platform. It would be nice to have board that supports more platforms at the same time (different compilers for the same target)

ivankravets:
Docs: http://docs.platformio.org/en/latest/platforms/creating_board.html

Usage
```
  "platforms": [
    "teensy", 
    "atmelavr",
    "..."
  ], 
```

Issue #479, opened 2016-01-27
"platform.json" manifest for development platform


ivankravets:
# Aims

We are very happy that PlatformIO community is growing up extremely fast. The last weeks have brought us a lot of requests to add new board or platform https://github.com/platformio/platformio/issues. This forces us to refactor PlatformIO 3.0 development platforms logic from the scratch. The main aim is to allow users, developers or even embedded vendors to create own "development platform" which can be easily installed using PlatformIO.

# Requirements

## Platform developer

Developer of "development platform" should provide the next information for PlatformIO Platform Manager:

* Generic platform information (name, version, etc)
* Package dependencies
* Platform build script
* Frameworks (supported by this platform) build scripts
* Boards list (only for embedded platforms)

**Package dependencies** should allow to use `semver` notation. For example, `"framework-mbed": "~1.2.0"`. In this case default PlatformIO Package Storage should be used. If package does not exist in PlatformIO Package Storage or developer wants to use own package, then allow to specify "direct download link" to it.

### PlatformIO Platforms Registry

Need to implement public Platforms Registry where will be located information about existing "development platforms". They will by used by http://docs.platformio.org/en/latest/userguide/platforms/cmd_search.html

# Using

# Platform is registered in Platforms Registry

It should look from the CLI as generic CLI for PlatformIO 2.0http://docs.platformio.org/en/latest/userguide/platforms/cmd_install.html

# Platform is not registered in Platforms Registry

Allow to install it directly using archive or folder which contains `platform.json`

```
platformio platform install file:///path/to/platform.zip
platformio platform install https://github.com/path/to/platform.zip
platformio platform install file:///path/to/platform/dir
```

Comments are welcome!

beriberikix:
Awesome! Would installing from git also be supported? NPM has a good [example for this](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies).

ivankravets:
@beriberikix I don't see any problems to not support CVS. However, need to understand that CVS clients are not installed on the all host OS. Maybe, some warnings to user?

ivankravets:
Examples: http://docs.platformio.org/en/feature-platformio-30/platforms/creating_platform.html#examples

PR #505, opened 2016-02-09
Support for multiple platforms in board definition. (Issue #456)

WisniewskiP:
I've added support for multiple platforms in board.json


ivankravets:
Thank you for pointing out this issue. Resolved in #479 for PlatformIO 3.0

PR #531, opened 2016-02-22
Fix incorrect download link.

NAzT:


ivankravets:
Thanks for the report! I've just updated links to the new storage.

PR #584, opened 2016-03-18
Fix resetmethod of ESPresso Lite 2.0

NAzT:


ivankravets:
Thank you for pointing out this issue. We have updated PlatformIO code and fixed this issue.

PR #597, opened 2016-03-26
Add link to instalation guide on ok1kvk.cz

skore13:


ivankravets:
Thanks a lot! I didn't merge this because it has typo and added within new commit.

PR #613, opened 2016-04-12
Fixed example configuration syntax

king-jam:


ivankravets:
Thank you for pointing out this issue. We have updated our docs.

PR #633, opened 2016-04-24
Add generic ATTiny45/85 support to board list

stonehippo:
Added basic support for generic ATTiny45/85 chips with USBTinyISP.

This is similar to the support for the Adafruit Trinket 8/16MHz boards, but is expected to be used with a generic chip, and a programmer such as the Tiny AVR Programmer from Sparkfun. My tests show that the Adfruit config works fine with that programmer, but adding the explicit board entries makes project config clearer and provides a place to drop additional config for these boards if required.

### DIFF
--- a/platformio/boards/attiny.json
+++ b/platformio/boards/attiny.json
@@ -1,0 +1,78 @@
+{
+  "attiny45internal1mhz": {
+    "build": {
+      "core": "arduino",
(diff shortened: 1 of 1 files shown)


ivankravets:
Thank you for pointing out this issue. We have added support for multiple ATTiny boards including ATTiny45/85. See http://docs.platformio.org/en/latest/history.html

PR #648, opened 2016-05-05
FIX inserted correct information for OpenBCI

pushtheworldllc:
Incorrect information for OpenBCI variant, led to a double entry for "OpenScope" under board selection dropdown.


ivankravets:
Thank you for pointing out this issue. We have updated the boards DB and fixed this issue.

PR #676, opened 2016-05-29
Msp430fr6989 fix

seryoni:
These look like copy paste issues


ivankravets:
Thank you for pointing out this issue. I've just updated configuration files.

PR #686, opened 2016-06-06
add --enable_upload functionality to "pio ci"

pvizeli:
I think it is a big benefit for library developers. It help to compile a examble or test file direct for uploading and don't need create a temp project anymore.

It also possible to compile a single i.e. ino file from a examble or other source to microcontroller without create a dumy project.


ivankravets:
Thanks a lot for your PR. See updated docs http://docs.platformio.org/en/latest/userguide/cmd_ci.html

Issue #289, opened 2015-09-21
Handle specific extra build flags from library.json


ivankravets:
Useful options from `platformio.ini` for [library.json](http://docs.platformio.org/en/latest/librarymanager/config.html)

* [build.flags](http://docs.platformio.org/en/latest/projectconf.html#build-flags)
* [build.unflags](http://docs.platformio.org/en/latest/projectconf.html#build-unflags)
* [build.srcFilter](http://docs.platformio.org/en/latest/projectconf.html#src-filter)
* [build.extraScript](http://docs.platformio.org/en/latest/projectconf.html#extra-script)

`library.json`
```json
{
  "name": "SomeLib",
   ...
  "build": {
    "flags": "-I target/n1/inc -DREV=13",
    "srcFilter": "+<*> -<test/> -<debug/>",
    "extraScript": "generate_headers.py"
  } 
}
```

PR #696, opened 2016-06-14
Apply BuildDependentLibraries to src_filter dirs

davidgraeff:
At the moment BuildDependentLibraries (scanning source files for lib
includes) is only executed for the _src_ directory, even if _src_filter_
brings in other directories. This few lines of code changes apply the
scanning to all additional dirs mentioned in src_filter with a "+".

## Use cases
Have targets in your platformio.ini which references additional source directories via the src_filter.

1) In my case i want to target an arduino and an esp8266 board. The core logic
    of my library is platform independant but some few files are board specific, so I need
    to pull them in via src_filter (which works great). This pull request allows mentioned
    libraries in those additional files to be handled like if the source files reside in "src/".

2) If you provide some examples with a library, at the moment it is difficult to provide the user
    a simple way to build an example. With this pull request the user just builds one of the
    targets in the platformio.ini which lists the example source files via src_filter. E.g.:

`[platformio]
src_dir = src/

[env:nodemcuv2lib]
platform = espressif
framework = arduino
board = nodemcuv2

[env:nodemcuv2Example1]
platform = espressif
framework = arduino
board = nodemcuv2
src_filter = +<\*> +<../examples/example1/src/*>`

### DIFF
--- a/platformio/builder/tools/platformio.py
+++ b/platformio/builder/tools/platformio.py
@@ -61,7 +61,7 @@ def BuildProgram(env):
             f.lower().strip() for f in env.get("FRAMEWORK", "").split(",")])
 
     # build dependent libs
-    deplibs = env.BuildDependentLibraries("$PROJECTSRC_DIR")
+    deplibs = env.BuildDependentLibraries("$PROJECTSRC_DIR", src_filter=env.get("SRC_FILTER"))
 
     # append specified LD_SCRIPT
     if ("LDSCRIPT_PATH" in env and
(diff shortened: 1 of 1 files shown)


ivankravets:
Thank you for this pull request.

1. You can write multiplatform code using different macros. PlatformIO passes these defines by default. For example, `-DESP8266`, `-DARDUINO_ARCH_AVR` and then `#ifdef... #endif`.
2. Take a look at [platformio ci](http://docs.platformio.org/en/latest/userguide/cmd_ci.html) command. You have the ability to run library example and pass the source code of library to build process.
3. We have a separate issue with extra options for the Library Builder. See #289

davidgraeff:
@ivankravets  Thanks for the fast response.

1) #289 is good thing indeed, will it be pio2 or be part of pio3?

2) platformio ci with the "-l" options is what I was looking for. On linux I made a symlink, that wasn't possible on windows machines. Is there already support in the atom based IDE for this workflow?

3) I still thing that "src" may either be an array with directories or src_filter added directories are handled
    in a similar fashion to the "src" dir.
    Sometimes the directory structure cannot be that simple. In my case one "platform" directory is a git submodule and I cannot easily add #ifdefs to the files, but the src_filter option would allow me to use this workflow.

ivankravets:
1. #289 is a part of PlatformIO 3.0. 
2. I don't recommend to mix "program source code" with "library source code". The aim of PlatformIO 3.0 is to make these components maximum isolated. As a result, we will introduce additional leverages to control build process for the libraries.

I propose to wait for the first PlatformIO 3.0beta and then we will discuss it.

In any case, thanks a lot for the PR and I'll not close it till final PlatformIO 3.0.

davidgraeff:
@ivankravets  ok then. i can live with my patched version of pio2 until pio3 is out

PR #706, opened 2016-06-26
Doc updates

jphollanti:
I just recently played with the library manager for the first time. With regards to registering your own library and from a perspective of a first time user, i thought these additions to docs/librarymanager/creating.rst would have been good to have in the documentation.


ivankravets:
Thanks a lot!!! 👍

PR #738, opened 2016-07-29
Update platformio.py

wuwx:
if has lib directory, and set ProcessUnFlags, will crash by message
TypeError: unhashable type: 'NodeList':

because lib's file a list, so call set function will be crashed


wuwx:
platformio.ini
```
extra_script = script/extra.py
```

script/extra.py
```
env.ProcessUnFlags([
    "-Wno-implicit-function-declaration",
])
```

ivankravets:
Thanks a lot! This is fixed in PlatformIO 3.0 branch. I've just merged PlatformIO 2.0 flags parser with PlatformIO 3.0

PR #744, opened 2016-08-09
Only do "please restart" message for boards that need it

palfrey:
In response to https://github.com/platformio/platformio/issues/709#issuecomment-238528332

I'll need to make a change for the Micro:bit board config to support this


ivankravets:
I think that good explanation in a few words could be a solution for this issue. No need to modify board manifests because we don't know about behaviour for other mbed-based boards.

ivankravets:
Let's keep the current text that we have (without config modificaiton)

PR #790, opened 2016-09-20
Add inheritance in platformio.ini file via includes

xoseperez:
The new dynamic variables feature in the platformio.ini file is cool but I think it's more flexible a inheritance approach like this one.

An "include" option is processed and expanded with all the options from the included section. It allows:
- Multiple inheritance (you can include more than one section in the same "include" option, comma-separated)
- Multilevel inheritance (a section can inherit from another section that inherits from a third one an so on)
- Recursion check (a section cannot inherit from another one that inherits from the first one)

An example of a platformio.ini file using this modification is:

```
[platformio]
env_default = test

[common]
platform = espressif8266
framework = arduino
lib_install = 19,64,89,408,549
build_flags = -DDEBUG_PORT=Serial
extra_script = pio_hooks.py

[ota]
upload_speed = 115200
upload_port = "mydevice.local"
upload_flags = --auth=mydevicepass --port 8266

[env:test]
include = common
board = nodemcuv2

[env:test-ota]
include = env:test,ota

[env:mydevice]
include = common
board = esp12e

[env:mydevice-ota]
include = env:mydevice,ota
```

### DIFF
--- a/platformio/exception.py
+++ b/platformio/exception.py
@@ -221,3 +221,12 @@ class CygwinEnvDetected(PlatformioException):
 
     MESSAGE = "PlatformIO does not work within Cygwin environment. "\
         "Use native Terminal instead."
+
+class UnknownIncludedSection(PlatformioException):
+
+    MESSAGE = "Unknown included section '{0}'"
+
+class RecursiveInclusion(PlatformioException):
+
+    MESSAGE = "Section '{0}' included recursively in '{1}'"
+
--- a/platformio/exception.py
+++ b/platformio/exception.py
@@ -222,11 +222,12 @@ class CygwinEnvDetected(PlatformioException):
     MESSAGE = "PlatformIO does not work within Cygwin environment. "\
         "Use native Terminal instead."
 
+
 class UnknownIncludedSection(PlatformioException):
 
     MESSAGE = "Unknown included section '{0}'"
 
+
 class RecursiveInclusion(PlatformioException):
 
     MESSAGE = "Section '{0}' included recursively in '{1}'"
-
--- a/platformio/util.py
+++ b/platformio/util.py
@@ -53,27 +53,28 @@ class ProjectConfig(ConfigParser):
 
     def items(self, section, **kwargs):
         items = {}
+
         try:
-            for option in ConfigParser.options(self, section):
-                if option == 'include':
-
-                    try:
-                        kwargs['previous'].append(section)
-                    except:
-                        kwargs['previous'] = [section]
-
-                    includes = self.get(section, 'include').split(',')
-                    for include in includes:
-                        if include in kwargs['previous']:
-                            raise exception.RecursiveInclusion(include, section)
-                        override = dict(self.items(include.strip(' '), **kwargs))
-                        items = self.merge(items, override)
-
-                else:
-                    items[option] = self.get(section, option)
-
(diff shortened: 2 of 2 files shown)


ivankravets:
Thanks a lot! I'll take a look at it later. The each  changes to `platformio.ini` should be reflected on PlatformIO IDE side.

xoseperez:
Can't see the problem with travis. Will this make it through?

ivankravets:
I don't like this `include` behavior. It looks messed according to C `#include` directive.

I'll contact here later. PlatformIO Core is very stable since 3.0 release. We don't have critical bugs. This allows us to focus on other tasks where community waits for them.

Thanks!

ivankravets:
Also, please don't close this PR. We will continue a discussion on it.

someburner:
+1

@ivankravets @xoseperez  - what needs to be done here? I can help perhaps.

This would make life so much easier for setting up multiple inis for projects that have/require many build flags.

For those who have not experience the glory of GNU Make's `include` directive.. you are missing out.

ivankravets:
@someburner have you seen http://docs.platformio.org/en/latest/projectconf.html#dynamic-variables?

someburner:
@ivankravets 

Yes, but what I want is to dynamically load other inis, not process input. With inis it's much easier to group defines. With this it appears i have to create some sort of script for each group of settings and pipe that into build_flags. I'm working with ~30 flags so the formatting here is actually quite important for organization. Not to mention editor integration where I can easily comment/uncomment things. 

To me it seems like piping should be handled separately from build flags. As I understand it, you can only pipe into build_flags? 

Im wary of using piped input into build flags as there are so many ways it could break down.. I mean, as stated it accepts literally *anything*, plus then your build flags are now platform dependent (i.e. version of shell, version of git, other env variables that can get set that affect that).

I didn't see a spec for the parser anywhere, or a guideline as to what kind of input it's expecting either (besides -D*something*, I mean more what constitutes as invalid input). It seems to not work when passing parenthesis if you are also trying to pass in other build flags from the ini (not sure if that's allowed either).

ivankravets:
> Yes, but what I want is to dynamically load other inis

This PR doesn't do that. It proposes another handling of dynamic vars within the same INI.

I understand what do you need. See http://docs.platformio.org/en/latest/projectconf.html#dynamic-build-flags

someburner:
If you can address my question about combining dynamic build_flags with other build_flags (is it supposed to be acceptable to do that?) then I can make an issue for it.

Still, it seems unreasonable to have to create a script for shell commands. And unnecessarily complicated if you just want to extend the INI parser to include another INI.

ivankravets:
> if you just want to extend the INI parser to include another INI.

Please open separate issue for that.

According to this PR, it will be closed soon because PIO Core 3.4 will be moved from ConfigParser to confobj that supports comments and etc.

ivankravets:
@someburner temporary, please take a look at http://docs.platformio.org/en/latest/projectconf.html#extra-script

ivankravets:
Thank you so much for your PR! I've refactored ProjectConfig and added support for global `[env]`. See 

- https://github.com/platformio/platformio-core/blob/develop/tests/test_projectconf.py
- https://github.com/platformio/platformio-core/blob/develop/platformio/project/config.py

Issue #830, opened 2016-11-18
Unable to include ssh user in lib_deps repository url


a2hill:
### Configuration

**Operating system**: macOS Seirra

**PlatformIO Version**: 3.1.0

### Description of problem
When including a git+ssh repository URL with the ssh username included the library installer will throw an AssertionError due to a version mismatch.

#### Steps to Reproduce

1. Add a lib_deps from a private repository (with username included): 
`git+ssh://git@gitlab.my-private-repo.com/a2hill/my-private-project.git#3.1.8`
2. Build the project
3. Receive error

### Actual Results
```
File "/Users/a2hill/.atom/packages/platformio-ide/penv/lib/python2.7/site-packages/platformio/managers/package.py", line 211, in _install_from_url
    pkg_dir = self._install_from_tmp_dir(pkg_dir, requirements)
File "/Users/a2hill/.atom/packages/platformio-ide/penv/lib/python2.7/site-packages/platformio/managers/package.py", line 249, in _install_from_tmp_dir
    assert tmpmanifest['version'] == requirements, mismatch_error
AssertionError: Package version 3.1.8 doesn't satisfy requirements gitlab.my-private-repo.com/a2hill/my-private-project.git#3.1.8
```

### Expected Results
Library should install correctly.

### If problems with PlatformIO Build System:

**The content of `platformio.ini`:**
```ini
[env:teensy31]
platform = teensy
board = teensy31
framework = arduino
lib_deps =
  git+ssh://git@gitlab.my-private-repo.com/a2hill/my-private-project.git#3.1.8
```

**Source file to reproduce issue:**
```cpp
N/A
```

### Additional info
This is important for privately hosted repository solutions (like GitLab) where a single 'git' user is used to control access to the repositories via public/private key authentication.

I know that URL's from lib_deps are sent directly to the platformio lib install utility. So after looking at the documentation I saw an option for `platformio lib [STORAGE_OPTIONS] install <name>@<version>`. To me it looks like the url parser is splitting on the '@' symbol and treating the second token as a version when in reality it is just the rest of the URL.

sbrinkmann:
@ivankravets it seems to be that the issue returned with the latest version if pio core.

ivankravets:
Should be fixed in previous releases. See https://github.com/platformio/platformio-core/blob/develop/tests/test_managers.py#L127

Issue #831, opened 2016-11-18
Unable to authenticate with username and password for private library dependencies


a2hill:
### Configuration

**Operating system**: macOS Seirra

**PlatformIO Version**: 3.1.0

### Description of problem
When including a lib_deps dependency using a git+https repository URL where authentication is required the library installer will fail.

#### Steps to Reproduce

1. Add a lib_deps from a private repository: 
`git+https://gitlab.my-private-repo.com/a2hill/my-private-project.git#3.1.8`
2. Build the project
3. Receive error

### Actual Results
```
Cloning into '/Users/a2hill/projects/my-private-thing/.piolibdeps/installing-byV_0o-package'...
fatal: could not read Username for 'https://gitlab.my-private-repo.com': Device not configured

CalledProcessError: Command '['git', 'clone', '--recursive', '--depth', '1', '--branch', '3.1.8', 'https://gitlab.my-private-repo.com/a2hill/my-private-project.git#3.1.8', '/Users/a2hill/projects/my-private-thing/.piolibdeps/installing-byV_0o-package']' returned non-zero exit status 128
```

### Expected Results
Library installer should prompt for Username and Password

### If problems with PlatformIO Build System:

**The content of `platformio.ini`:**
```ini
[env:teensy31]
platform = teensy
board = teensy31
framework = arduino
lib_deps =
  git+https://gitlab.my-private-repo.com/a2hill/my-private-project.git#3.1.8
```

**Source file to reproduce issue:**
```
N/A
```

### Additional info
This is important for privately hosted repositories where authentication is required. Additionally if the username is included like `git+https://a2hill@gitlab.my-private-repo.com` the error will read `could not read Password for...`. You will also run into the issue #830 .

ivankravets:
What is your git version? My version asks me about username/password
```
pio run -d 
[Fri Dec  2 21:05:14 2016] Processing teensy31 (build_flags: -O  -D TEENSY31=1 -UUSB_SERIAL -DUSB_SERIAL_HID -DLAYOUT_US_ENGLISH -I./src, lib_deps: git+https://gitlab.com/user/project.git#0.6.0, board_f_cpu: 96000000, platform: teensy, board: teensy31, framework: arduino)
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
LibraryManager: Installing project
git version 2.9.3 (Apple Git-75)
Cloning into '.../.piolibdeps/installing-E1qs8l-package'...
[...]

PR #836, opened 2016-11-22
Allow git ssh user to be included with lib_deps url

a2hill:
A patch to fix issue #830 

This patch takes the existing conditionals, which check for an '@' character indicating a required version of the given dependency, and adds an additional check for the 'git' username when used in conjunction with a protocol like git+ssh. 

This fixes the current behavior where a URL in the form of 
```
git+ssh://git@my-private-repo.com/some-user/some-project.git
```
 would be parsed as a dependency having a required version of `my-private-repo.com/some-user/some-project.git` which causes an exception during installation.

I have also reorganized the conditionals into a list for readability.

### DIFF
--- a/platformio/managers/package.py
+++ b/platformio/managers/package.py
@@ -297,12 +297,13 @@ class BasePkgManager(PkgRepoMixin, PkgInstallerMixin):
     def parse_pkg_name(  # pylint: disable=too-many-branches
             text, requirements=None):
         text = str(text)
-        if not requirements and "@" in text and not text.startswith("git@"):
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks!

Does it resolve both issues: #830 and #831?

a2hill:
Unfortunately no, this just resolves #830. I might look into #831 but realistically that was only an issue for me when I was trying to use it as a workaround to #830

ivankravets:
I'll take a look on it tomorrow.

PR #844, opened 2016-12-01
Feature/sort serialports

zeroflow:
Simply sort the list of serial ports from util.get_serialports

**Reason:**
When no HwId is given in the board configuration the autodetect function will simply use the last port returned from get_serialports()
At least on windows, the returned list is not guranteed to be in any specific order.
To make uploading more predictable, I suggest sorting the ports to have a strict ascending order.

**Background:**
On my PC I have a constantly connected Arduino Leonardo (COM1) and e.g. a ESP8266 board (COM6).
When using autodetect, it often returns the list [COM6, COM1] and tries to upload on COM1.
Even when setting the Leonardo to COM265 it still returns [COM6, COM265].
My change allows to set constantly attached boards to low COM-Ports which will not be selected by autodetect when a higher-number COM port is available.

Sorry for the not-yet-signed CLA, CLAHub is throwing me errors when trying to sign it

### DIFF
--- a/platformio/util.py
+++ b/platformio/util.py
@@ -378,6 +378,11 @@ def get_serialports(filter_hwid=False):
 
     if filter_hwid:
         return result
+        
+    if system() == "Windows":
+        result = sorted(result, key=lambda x:int(x["port"][3:]) )
+    else:
+        result = sorted(result, key=lambda x:x["port"])
 
     # fix for PySerial
     if not result and platform.system() == "Darwin":
--- a/platformio/util.py
+++ b/platformio/util.py
@@ -378,11 +378,11 @@ def get_serialports(filter_hwid=False):
 
     if filter_hwid:
         return result
-        
-    if system() == "Windows":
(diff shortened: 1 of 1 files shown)


ivankravets:
Have you heard about http://docs.platformio.org/en/stable/projectconf.html#upload-port?

zeroflow:
Yes, since it's a well documented feature.

This upload_port works but for example when uploading the code to multiple boards I often get different assigned COM-Ports.

I've Seen that there is an open issue to support multiple ports and maybe Wildcards. That part may be a better place to host such functionality.

ivankravets:
Using PlatformIO 3.2 you can clone existing board manfiest to `project/boards/*` and correct hardware VID/PID accordinng your USB/UARTs.

In any case, we have a lot of instruments that allow to control upload ports dynamically. Take a look at http://docs.platformio.org/en/stable/projectconf.html#extra-script

PR #960, opened 2017-05-10
Add templates for VS Code

dymo:
Add templates for Visual Studio Code support.


dymo:
I know very little about Python, but it looks like the build errors are not related to my commits?

PR #979, opened 2017-06-08
Add udev rule for permissions mode for USBasp V2.0

code-sur:


ivankravets:
Thank you very much for PR!

PR #995, opened 2017-07-07
Escape spaces in task command for VSCode Task Runner

jclaessens97:
Refering to https://community.platformio.org/t/space-in-path-causes-error/2127 and just escaping the " character doesn't work, I think this solves the problem.


ivankravets:
There is no file `platformio/ide/tpls/vscode/.vscode/tasks.json.tpl`. Please upgrade to PlatformIO IDE for VSCode >= 0.5.0

PR #1051, opened 2017-09-08
Avoid Dwengo board not being detected after force reset

itkovian:
Uploading to my Dwengo board always fails: the board is not found.

Increasing the delay after the reset command fixes this.

### DIFF
--- a/platformio/builder/tools/pioupload.py
+++ b/platformio/builder/tools/pioupload.py
@@ -48,7 +48,7 @@ def TouchSerialPort(env, port, baudrate):
         s.close()
     except:  # pylint: disable=W0702
         pass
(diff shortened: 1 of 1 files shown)


ivankravets:
Could you restore to `0.4` value and extend time out [here](https://github.com/platformio/platformio-core/blob/674aa5c4d85344f61b9851b28c9b2abc8b8742f9/platformio/builder/tools/pioupload.py#L60). Please change `5` to `6`, for example. Does it work?

itkovian:
No, but I did find out that the WaitForNewSerialPort was not called in the dwenguino case, see https://github.com/platformio/platform-atmelavr/pull/70

PR #1287, opened 2018-01-09
Additional fix to 7354515

dbolotin:
The same fix as in commit 7354515 by @ivankravets, plus:
* correct handling of macro-functions with concatenation (`##`) operation
* correct behaviour for spaces in macros. Without quotes around `-D` option's argument I experienced errors like:
  ```
  Problems were encountered while collecting compiler information:
  	xtensa-lx106-elf-g++: error: long: No such file or directory
  	xtensa-lx106-elf-g++: error: unsigned: No such file or directory
  	xtensa-lx106-elf-g++: error: int: No such file or directory
  ```
  for defines like `add_definitions(-D__UINT_LEAST64_TYPE__=long long unsigned int)`
  "idea" of `-D...` -> `-D'...'` taken form [here](https://gcc.gnu.org/onlinedocs/cpp/Invocation.html).

This patch:
* Tested on CLion 2017.2.3: IDE imports project without warnings or errors
* Verified correct macro expansion by IDE for macros with hashes (i.e. `__UINT32_C(12)` expands to `12UL`)

### DIFF
--- a/platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl
+++ b/platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl
@@ -8,7 +8,7 @@ SET(CMAKE_C_FLAGS_DISTRIBUTION "{{cc_flags}}")
 set(CMAKE_CXX_STANDARD 11)
 
 % for define in defines:
-add_definitions(-D'{{!re.sub(r"([()#])",r"\\\1",define)}}')
+add_definitions(-D'{{!re.sub(r"([\"()#])",r"\\\1",define)}}')
 % end
 
 % for include in includes:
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks a lot for PR!

ivankravets:
> Tested on CLion 2017.2.3:

How did you test it? 😊 This PR raises "NameError: name 're' is not defined" and not all items from regular expression are escaped. I fixed that in new commit  https://github.com/platformio/platformio-core/commit/61eb989eddd40465e58e7f142f154b1465432dba

Please re-run commands below and retest. Thanks!
```
pio upgrade --dev
pio init --ide clion
```

dbolotin:
Yes! Sorry. 🤦‍♂️  I tested it by in-place changing files in my local platformio installation and then forgot to transfer the `import` statement into PR.

I tested current development version, everything works, CLion flawlessly imports the project!

Thanks for quick merge!

legap:
Thank you both for this fix!

I was confronted with the same problem and this PR finally solved those CMake errors.
CLion works again as expected after an upgrade with --dev :thumbsup:

PR #1290, opened 2018-01-12
Fix: pio init generate wrong DEFINES lines for QtCreator

hongquan:
It created the line
```pro
DEFINES += "__INT64_C(c)=c ## LL"
```
which is seen as wrong syntax.

The fix helps generate line:
```pro
DEFINES += "__INT64_C(c)=c"
```


ivankravets:
Please run `pio upgrade --dev` and re-init project with `pio init --ide qtcreator`.

Does it work now?

hongquan:
@ivankravets It works!

hongquan:
@ivankravets No, actually, it causes another bug. Will make another report.

Issue #895, opened 2017-02-03
Make PlatformIO Python 3 compatible


StefanD986:
### Configuration

**Operating system**:
Windows and Mac

**PlatformIO Version** (`platformio --version`):
PlatformIO, version 3.2.1

### Description of problem
PlatformIO is not compatible to Python 3 (as stated on the website). It would be nice and would ease the install process if PlatformIO would also work with Python 3.

ivankravets:
PlatformIO's Build System depends on [SCons](http://scons.org) tool that doesn't support Python 3. Please note that SCons team works in this way.

skorokithakis:
@ivankravets How hard would it be to remove this dependency, or replace it with something else?

Would it be feasible to write an adapter module that called out to SCons from Python 3? Although that would be a hassle when installing...

Python 2 is quickly becoming irrelevant, and I'm finding myself having to keep a Python 2 env around just for PlatformIO. I think this should be a higher priority, as PlatformIO risks finding itself in a position where it can't be easily installed soon (with OSes moving to Py3).

I would be willing to contribute some code to the migration to Py3.

ivankravets:
>  PlatformIO risks finding itself in a position where it can't be easily installed soon (with OSes moving to Py3).

Python 2 is a stable and supported branch. I think, minimum 5 years all major OSs will include Python 2 distribution by default (as they now do). Hundred OS-dependent libraries and software are written in Python 2: macOS, Linux, BSD.

Nevertheless, SCons Team [hardly working on the first Python3 compatible release](https://bitbucket.org/scons/scons/commits/all).

- https://bitbucket.org/scons/scons/src/9ef88cce4c52e65083beba3f416d45a494855693/src/CHANGES.txt?at=default&fileviewer=file-view-default

tehnorm:
Just noticed that it looks like SCons 3.0.0 just released support for python 3.5+!

http://scons.org/scons-300-is-available.html

ivankravets:
Upgraded to SCons 3.0 in PlatformIo Core 3.5.2 ( https://github.com/platformio/platformio-core/issues/1253 ). Now, need to upgrade PIO Core to Python 3.

mintisan:
How can I install the specific version for Python3?

It seems to be not OK, if I use `pip3 install platformio`.

```
➜  ~ pip3 install platformio
Collecting platformio
[...]

PR #1365, opened 2018-02-14
feat(py3): add Python 3 support (backwards compatible with Python 2.7)

cfobel:
See #895.

Note: platforms/packages containing Python code, e.g.,
[`platform-atmelavr`][1], still need to be updated to support Python 3.

The [`python-modernize`][2] can help with this considerably, e.g.:

    python-modernize . -w

will automatically use the [`six`][3] module to apply numerous
transformations required for Python 2/3 compatability.  The
[`python-modernize`][2] tool was used for most Python 3 updates to
`platformio-core` introduced by this merge, with the notable exception of
unicode string handling, which required manual intervention (see below).

Unicode issues
--------------

In Python 3, **all** string literals (i.e., `str` objects) are **unicode** -
**there is NO `unicode` type** in Python 3.  In Python 3, the **`bytes` type**
is equivalent to **`str` in Python 2**.  Note that in Python 2 the `bytes` type
may be used interchangeably with the `str` type.

However, in Python 3 many library functions, e.g., `hashlib.md5.update` require
`bytes` objects; **NOT `str` objects**.  Thus, to work properly with Python 3,
code ported from Python 2 must be updated to **explicitly decode** `str`
unicode string objects to `bytes`, e.g., `my_string.encode('utf8')`.

Furthermore, some Python 3 library functions now **return** `bytes` objects,
e.g., `Popen.communicate`.  Python 2 code interpreting the output of such
functions as `str` string objects **MUST** be updated to explicitly encode the
output to unicode strings to work properly in Python 3, e.g.:

```python
process = subprocess.Popen(...)

stdout, stderr = process.communicate()

print(stdout.decode('utf8'))
```

[1]: https://github.com/platformio/platform-atmelavr.git
[2]: https://github.com/python-modernize/python-modernize
[3]: https://pythonhosted.org/six/

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,8 @@ PlatformIO 3.0
 3.5.2 (2018-??-??)
 ~~~~~~~~~~~~~~~~~~
 
+* Upgraded build system to `SCons <http://scons.org>`__ 3.0
+  (`issue #895 <https://github.com/platformio/platformio-core/issues/895>`_)
 * Added aliases (off, light, strict) for
   `LDF Compatibility Mode <http://docs.platformio.org/page/librarymanager/ldf.html>`__
 * Show device system information (MCU, Frequency, RAM, Flash, Debugging tools)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -508,7 +508,7 @@ PlatformIO 2.0
 * Added support for `emonPi <https://github.com/openenergymonitor/emonpi>`__,
   the OpenEnergyMonitor system
   (`issue #687 <https://github.com/platformio/platformio-core/issues/687>`_)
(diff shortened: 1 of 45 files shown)


ivankravets:
Thank you so much for this great step ahead! I'll back this PR soon. We have related PIO Plus project which is written in Python 2.7. Need to port them together.

cfobel:
@ivankravets No problem!  Thank you for all of your hard work on developing the PIO ecosystem!

Cabalist:
I'd really like to see this enhancement merged.  Is there anything blocking that I can help with?

ivankravets:
Thank you so much for your PR. Actually, we don't need a lot of these changes by modernizer. See a final commit which introduces full support for Python3.

Please re-test with `pio upgrade --dev`. It should be PlatformIO 4.0.

Does it work for you?

PR #1429, opened 2018-03-10
Fix CLion not recognizing includes in lib and .piolibdeps

JackDn:
This PR fixes the problem discussed in [this post](https://community.platformio.org/t/clion-fails-to-find-include-files-in-lib-folder/3247) by adding source files from `lib` and `.piolibdeps` to `SRC_LIST`.


ivankravets:
Thank you so much!

DL6AKU:
Thanks =)

PR #1584, opened 2018-04-20
shutil.move() instead of os.rename()

lobradov:
os.rename only works if src and dst are on the same filesystem, and will fail otherwise.
shutil.move() handles these cases better:
https://www.pythoncentral.io/how-to-rename-move-a-file-in-python/


ivankravets:
I did some changes to formatting and pushed to develop branch. Let's see how it will work.

Thanks!

PR #1634, opened 2018-05-24
Don't add ".vscode/xxx" to .gitignore if .vscode is already ignored.

hongquan:
Use case:
 My _.gitignore_ already has `.vscode` entry, but PlatformIO insists on adding `.vscode/c_cpp_properties.json` and alike to my file.


ivankravets:
Resolved in https://github.com/platformio/platformio-core/commit/22ceae014919a427b97a6a51d39c71b1f82f3b55

PR #1699, opened 2018-06-21
Add udev rule for Maple Leaf board

dvv:


dvv:
I doubt udev rules are under CI, can't figure out the checks failure reason

ivankravets:
See http://docs.platformio.org/en/latest/projectconf/advanced_scripting.html#custom-upload-tool

ivankravets:
Thanks!

PR #1710, opened 2018-06-28
please add a new boards STM32F103ZET6

EmiyaNing:
please add a new boards STM32F103ZET6


ivankravets:
Please file a feature request here https://github.com/platformio/platform-ststm32/issues

PR #1873, opened 2018-10-09
Solved issues with vim whitespaces in paths

JoseAntonioTorre:
When a library has whitespaces in the name the path
will contain whitespace too. When vim tries to decode
path it will fail.

This fix wrap each path with quotes.

### DIFF
--- a/platformio/ide/tpls/vim/.clang_complete.tpl
+++ b/platformio/ide/tpls/vim/.clang_complete.tpl
@@ -1,5 +1,5 @@
 % for include in includes:
--I{{include}}
+-I"{{include}}"
 % end
 % for define in defines:
 -D{{!define}}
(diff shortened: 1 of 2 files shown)


ivankravets:
Thank you so much for PR! Do you know which VIM's packages use `.gcc-flags.json` and `.clang_complete`? I see that this works very well for Atom.

JoseAntonioTorre:
> Thank you so much for PR! Do you know which VIM's packages use `.gcc-flags.json` and `.clang_complete`? I see that this works very well for Atom.
Hi!
I am using:
* meck/ale-platformio: this one uses .gcc-flags.json to configure clang in order to lint the code. https://github.com/meck/ale-platformio/projects
* Rip-Rip/clang_complete: this uses .clang_complete in order to configure clang to provide a source to deoplete plugin.

If you want to test the problem and the fix you can create a new project with: pio init --ide vim -b uno and you will see that vim works (if you don't have any whitespace in the path).
Then install a library with a space in it's name, for example: https://platformio.org/lib/show/5354/Grove%20I2C%20Motor%20Driver%20v1.3
and regenerate the files with "pio init --ide vim -b uno". 
Now the linter has gone and the completion will work until the library path (libraries above will work).

Let me know if you have any problem trying to reproduce.
Thanks.

JoseAntonioTorre:
@ivankravets Could you try it?

ivankravets:
thanks!

PR #2143, opened 2019-02-18
InoToCpp - Ignoring spaces, tabs, .. for prototype detection

vossim:
Ignoring spaces, tabs, etc for prototype detection in InoToCPPConverter.


vossim:
Handling whitespaces isn't done correctly, but with this patch classes are broken (for example in the basic test for ino2cpp)

PR #2264, opened 2019-03-29
Surround ld script path in quotation marks

robinvalk:
While following [this tutorial](https://docs.platformio.org/en/latest/tutorials/ststm32/stm32cube_debugging_unit_testing.html#tutorial-stm32cube-debugging-unit-testing) I got a `No such file or directory` error during the build. After building with the verbose tag I saw that the path was not surrounded in quotation marks. This resulted in gcc not finding the file as the path contained a space.

After locally adding the marks in the script the build succeeded. I don't think it is the responsibility for the platforms to provide the `LDSCRIPT_PATH` environment variable already enclosed is it?

### DIFF
--- a/platformio/builder/tools/platformio.py
+++ b/platformio/builder/tools/platformio.py
@@ -121,7 +121,7 @@ def BuildProgram(env):
     # append into the beginning a main LD script
     if (env.get("LDSCRIPT_PATH")
             and not any("-Wl,-T" in f for f in env['LINKFLAGS'])):
-        env.Prepend(LINKFLAGS=["-T", "$LDSCRIPT_PATH"])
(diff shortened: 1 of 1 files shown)


Tahsinbozkurt:
tanks

ivankravets:
This is an issue of dev/platform. See fix here https://github.com/platformio/platform-ststm32/issues/229

ivankravets:
In any case, thank you so much for your PR!

robinvalk:
Ah, so it is a platform issue ^^. Thanks Ivan!

muhammedtapis:
how do i find id script path in which file ? thanks

PR #2676, opened 2019-06-16
Fix Teensy serial lockout

pfeerick:
Fixes `[Errno 16] Device or resource busy: '/dev/ttyACM0` issue on Teensy 3.2 and 3.5 (at minimum) due to ModemManager probing the devices and making them unavailable. Updated rules sourced from current https://www.pjrc.com/teensy/49-teensy.rules configuration.

Came across the issue due to this forum post: https://community.platformio.org/t/platformio-difficult-to-configure/8189/2


ivankravets:
Thanks 🙏

PR #2700, opened 2019-06-25
Missing parentheses and depreciated syntax

pfeerick:
Shouldn't have been able to break this the way I did... but I did (by-product of trying to coerce an embedded python install to work with platformio on windows). python3 hurdles. 

```
  File "get-platformio.py", line 93
    print r['out']
          ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(r['out'])?
```

and 

```
  File "get-platformio.py", line 146
    except Exception, e:
                    ^
SyntaxError: invalid syntax
```


ivankravets:
Thanks!

PR #2714, opened 2019-06-27
Fix for tpls/eclipse/.cproject.tpl

awachtler:
In the current version of this template, all build targets result in a command line

   pio -c -f eclipse debug run target <mytarget>

The commit fixes this to be

   pio -c -f eclipse run target <mytarget>

See also discussion in forum:
   https://community.platformio.org/t/pio-4-0-0b3-potential-bug-in-cprojet-tpl/8390


ivankravets:
Thank you so much! This issue is only reproducible when build configuration is set to "Debug".

PR #2916, opened 2019-08-17
Clean-up error handling in 'run' command

mcspr:
- create `handle_legacy_build_dir()` and move '.pioenvs' clean-up code there
- make sha1() py3-compatible. otherwise, it will trigger bogus warning about the need to manually remove build directory. the real exception is:
```python
  File "/home/builder/git/platformio-core/platformio/project/helpers.py", line 170, in compute_project_checksum
    checksum = sha1(__version__)
TypeError: Unicode-objects must be encoded before hashing
```
- show original exception via fs.rmtree onerror handler
for example:
```python
[Errno 13] Permission denied: '/home/builder/project/.pioenvs'
Please manually remove `/home/builder/project/.pioenvs` to avoid build issues
```
instead of
```python
[Errno 21] Is a directory: '/home/builder/project/.pioenvs'
Please manually remove the file `/home/builder/project/.pioenvs`
```
which is not useful at all

### DIFF
--- a/platformio/commands/run/command.py
+++ b/platformio/commands/run/command.py
@@ -23,6 +23,7 @@ from platformio import exception, fs
 from platformio.commands.device import device_monitor as cmd_device_monitor
 from platformio.commands.run.helpers import (clean_build_dir,
                                              handle_legacy_libdeps,
+                                             handle_legacy_build_dir,
                                              print_summary)
 from platformio.commands.run.processor import EnvironmentProcessor
 from platformio.project.config import ProjectConfig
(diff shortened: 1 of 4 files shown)


ivankravets:
Thanks!!! Please re-test with `pio upgrade --dev`.

mcspr:
Cool.

What about shadowed exception from the fs.rmtree / os.rmtree onerror? While .pioenvs-specific message was removed, exception text already describes what is wrong

ivankravets:
Could explain in details?

mcspr:
So, the example above is one of the issues - the error show is the one from the `os.remove(...)` call, not the original error. My guess, the reason for that handler is the same as for pip here?
https://github.com/pypa/pip/blob/933280308968b59b24389f85186021d3dfbd7051/src/pip/_internal/utils/misc.py#L179-L191

We expect the method to succeed after chmod. Two problems, however:
- while onerror's first argument is the original func that failed, os.remove is always used. it will not work on directories, for one, and show the "is a directory" exception as the failure cause. The way I did it here is wrong though... using onerror's first argument is a more clear way to issue the error again (or something different)
for example: https://github.com/python/cpython/blob/24fe46081be3d1c01b3d21cb39bc3492ab4485a3/Lib/shutil.py#L596
- `os.chmod(file, MODE)` is better to rewrite as `os.chmod(file, os.stat(file).mode | MODE)`, because we end up with strange `d-w-------` for directories. "Permission denied" is shown by doing `ls` in shell, for example.
underlying dir can have some other users files, or Windows can lock up the directory by doing things like `notepad.exe > now_we_are_locked.txt`

ivankravets:
I see. Thanks so much for this valuable report! Fixed in https://github.com/platformio/platformio-core/commit/bf59dda01bbabf6efe248773bfc347bd2bd6ea09

I tested on Windows VM and it works now. I would be thankful if you re-test too.

mcspr:
Thanks! It does show proper warnings and handles directories deletion for me too.

I was trying to break permissions some more... the only case where secondary error is still vague and re-raising the original one still makes sense, is when dir file mode is `000` / `d---------` (which is manually done, and probably wont happen in the real world cases)
It shows ` open() missing required argument 'flags'` when it was `[Errno 13] Permission denied` originally.

PR #2944, opened 2019-08-25
Clion integration, resolves #2824

### LINKED ISSUE #2824: Feature request : Better integration of environments in CLion
Hi. I love platformio and have been using it for two years with CLion without a hitch. Recently I have started using multiple environments but they do not integrate at all in CLion :
 - Cannot run individual environments from the IDE
 - Code completion features are based on the first environment only

Those issues forced me to add a dummy environment, write some custom CMake and have some manual setup after each `pio init`, even though I felt that this could be integrated in platformio.

In the process of solving this issues for my project, I found ways around them which seemed very portable and thought that they might help enhance platformio's integration with CLion.

I forked the repo, tried some things around and I managed to solve the above-mentioned issues. Should I submit a PR ?

Teo-CD:
Better CLion integration using the new access to ProjectConfig from ProjectGenerator ( 1dc15326c9d5d3e894e20d5ed2dfc89d4c9358e1 ).

Thanks to the changes this was much cleaner and faster to do. I would have preferred not to use `load_project_ide_data()` to get the full list of includes and defines but I could not figure how to do it easily just with `config`.

Currently the build profile allowing to run every environment (previous behaviour) is the last on the list, would you prefer it to be the default  ?

### DIFF
--- a/platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl
+++ b/platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl
@@ -5,7 +5,7 @@
 # please create `CMakeListsUser.txt` in the root of project.
 # The `CMakeListsUser.txt` will not be overwritten by PlatformIO.
 
-%from platformio.project.helpers import (get_project_dir,load_project_ide_data)
+%from platformio.project.helpers import (load_project_ide_data)
 %
 % import re
 %
@@ -23,7 +23,6 @@
 % end
 %
 % envs = config.envs()
-% project_dir = get_project_dir()
 
 % if len(envs) > 1:
 set(CMAKE_CONFIGURATION_TYPES "{{ ";".join(envs) }}" CACHE STRING "" FORCE)
@@ -56,7 +55,7 @@ if (CMAKE_BUILD_TYPE MATCHES {{ env }})
 %   end
 
 %   for include in items["includes"]:
-    include_directories("{{ _normalize_path(include) }}")
+    include_directories("{{ _normalize_path(to_unix_path(include)) }}")
 %   end
 endif()
 % end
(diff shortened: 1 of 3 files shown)


platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl
```diff
@@ -5,6 +5,8 @@
 # please create `CMakeListsUser.txt` in the root of project.
 # The `CMakeListsUser.txt` will not be overwritten by PlatformIO.
 
+%from platformio.project.helpers import (get_project_dir,load_project_ide_data)
```
ivankravets: > get_project_dir

There is exported `project_dir` variable which is already normalized for Windows OS. See 
> https://github.com/platformio/platformio-core/blob/develop/platformio/ide/projectgenerator.py#L73

platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl
```diff
@@ -37,12 +47,16 @@ SET(CMAKE_C_STANDARD {{ cc_stds[-1] }})
...
+if (CMAKE_BUILD_TYPE MATCHES {{ env }})
+%   items = load_project_ide_data(get_project_dir(),env)
+%   for define in items["defines"]:
+    add_definitions(-D'{{!re.sub(r"([\"\(\)#])", r"\\\1", define)}}')
+%   end
 
-% for include in includes:
-include_directories("{{ _normalize_path(include) }}")
+%   for include in items["includes"]:
+    include_directories("{{ _normalize_path(include) }}")
```
ivankravets: As you use raw includes from build env, need to convert path to unix style, CMake does not like `\\` Windows slahes. So, please use ``{{  _normalize_path(to_unix_path(include)) }}``.

Teo-CD:
Totally slipped my mind. Now fixed !

ivankravets:
>  I would have preferred not to use load_project_ide_data() to get the full list of includes and defines but I could not figure how to do it easily just with config.

Resolved in:
- https://github.com/platformio/platformio-core/commit/83bf34fb7766894c8873e809e247de537e5d4b30

See updated file https://github.com/platformio/platformio-core/blob/develop/platformio/ide/tpls/clion/CMakeListsPrivate.txt.tpl#L60

PR #2966, opened 2019-08-31
Stop ModemManager corrupting Arduino uploads

pfeerick:
On boards like the Arduino Micro, when in bootloader mode it appears ModemManager interferes with the programming process and result in a catastrophic failure with no end of different errors including, but not limited to:
```
error: programmer did not respond to command: write block
error: butterfly programmer uses avr_write_page() but does not provide a cmd() method.
error: programmer did not respond to command: set addr
```
After this, the device could appear to be completely non-functional, refusing to enumerate or appear for programming, but thankfully a double-reset will usually recover it, but the underlying ModemManager issue will still prevent successful programming. Hence the additional rules. 

This affects not only PlatformIO, but also the Arduino IDE (on linux). 

For my device, I only needed the first of two lines edited, but might as well do both as it will likely cause problems for all the range. I have not tested this outside of my own development box, but ModemManager is the cause of no end of Arduino upload issues on linux.

### DIFF
--- a/scripts/99-platformio-udev.rules
+++ b/scripts/99-platformio-udev.rules
@@ -40,8 +40,8 @@ SUBSYSTEMS=="usb", ATTRS{idVendor}=="067b", ATTRS{idProduct}=="2303", MODE:="066
 SUBSYSTEMS=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="7523", MODE:="0666"
 
 # Arduino boards
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="2341", ATTRS{idProduct}=="[08][02]*", MODE:="0666"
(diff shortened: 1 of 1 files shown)


pfeerick:
And now that I think of it, I now need to see if this is also what is causing issues with the STM32 and DFU uploads on linux...

ivankravets:
Should we add this for other rules?

pfeerick:
Since ModemManager is only really supposed to interact with...
> mobile broadband (2G/3G/4G) devices and connections

... and these are not those... so probably should, the need for this stems from a change in behaviour in ModemManager that doesn't sound like it's going away. I've just only been making changes when I've encountered problems and hadn't detected any unforeseen side effects. 

If you want me to update all the entries, I'll go ahead and update the PR.

ivankravets:
Good, let's wait if someone reports any issues. Thanks!

PR #3018, opened 2019-09-15
Run client child process with a pty

benmezger:
Setting `usePTY` allows better interaction with GDB. `usePTY` is recommended when programs primarily interact with users via a terminal.


benmezger:
Do I have to close the PR and reopen to get clahub to pass? (signed the agreement after doing the PR).

ivankravets:
Thanks!

1) `ValueError: The usePTY parameter is not supported on Windows.`. It's not supported on Windows.
2) Do you have any issues on Unix without `usePTY=True`?

PR #3026, opened 2019-09-16
UTF8 decoding should ignore invalid characters coming from the serial connection

ThomasBleijendaal:
Some boards, like ESP32 based boards, give some unintelligible data when connecting to them via Serial. This is sometimes data that is send with the wrong baud rate (hard baked into the boot loader), or something else. It's hard to prevent this from happening. When a build is uploaded to the ESP board for unit testing, the decoding of the incoming stream should not fail the test due to some garbled content. Since the read data is validated on line 95, any garbage is automatically ignored and only outputted to the console.


ivankravets:
Thanks, Thomas! 👍

PR #3032, opened 2019-09-18
udev: Add GD32V DFU Bootloader

adlerweb:
Used for example on [Sipeed Logan Nano](https://docs.platformio.org/page/boards/gd32v/sipeed-longan-nano.html)


ivankravets:
Thanks!

Issue #2990, opened 2019-09-12
Feature Request: Generate compile_commands.json


Knowledge91:
I am using emacs with `lsp-mode` and *clangd* language server. For code completion the language server need an **compile_commands.json** file, which can be produced by cmake.
- [compile_commands.json definition](https://clang.llvm.org/docs/JSONCompilationDatabase.html)
- [How to produce compile_commands.json with cmake](https://clang.llvm.org/extra/clangd/Installation.html)

I would love to see a command similar to `pio init --ide emacs` that produces a compile_commands.json instead of a .clang_complete file, which I need for Intellisense.

ivankravets:
See https://github.com/platformio/platformio-core/commit/7a07a2e63eb51f8e403ecaf0fb2e4029d7b54744

Could you re-test with `pio upgrade --dev`?

Knowledge91:
I did and noticed that you produce an additional .ccls file, which is used by the ccls language server. Unfortunately cland, the language server of the llvm developers, uses a ‘.compile_commands.json’ file

ivankravets:
How are you going to use clang with `compile_commands.json` when it's not compatible with GCC flags and other architecture features? It will fail, just try it for AVR.

Knowledge91:
Haha. Actually I just want the intellisense to work with cland. For compilation I just the pio pipeline.

ivankravets:
Yes, but clang C/C++ Preprocessor will not understand your GCC code. You actually need clang_complete and we generate it https://github.com/platformio/platformio-core/blob/develop/platformio/ide/tpls/emacs/.clang_complete.tpl

Knowledge91:
You don’t get me. The .clang_complete file is not read by the cland language server, which is the default language server for the emacs lsp-mode. It’s like the .ccls file for the ccls language server 😉
Anyways thanks for your quick responses. Keep up the good work.

mnrce:
@ivankravets:
Hi, can this issue be re-opened?
Each C++ language servers support its own compile-flag file: 
- For `ccls`, it is [.ccls](https://github.com/MaskRay/ccls/wiki/Project-Setup#ccls-file) (which PIO generates)
- For `clangd`, it is [compile_flags.txt](https://clang.llvm.org/extra/clangd/Installation.html#compile-flags-txt)
- For `cquery`, it is [.cquery](https://github.com/cquery-project/cquery/wiki/.cquery)

[...]

PR #3051, opened 2019-09-26
CMake IDE/TPL added

xeno010:
This is very helpful when working with cmake without a specific IDE.

```
# Create new cmake project
platformio init --ide cmake --board <ID>
```
11 predefined targets for building
```
PLATFORMIO_BUILD - Build project without auto-uploading
PLATFORMIO_BUILD_VERBOSE - Build project without auto-uploading in verbose mode
PLATFORMIO_UPLOAD - Build and upload (if no errors)
PLATFORMIO_CLEAN - Clean compiled objects
PLATFORMIO_MONITOR - Device monitor platformio device monitor
PLATFORMIO_TEST - PIO Unit Testing
PLATFORMIO_PROGRAM - Build and upload using external programmer (if no errors), see Upload using Programmer
PLATFORMIO_UPLOADFS - Upload files to file system SPIFFS, see Uploading files to file system SPIFFS
PLATFORMIO_UPDATE - Update installed platforms and libraries via platformio update
PLATFORMIO_REBUILD_PROJECT_INDEX - Rebuild C/C++ Index for the Project. Allows one to fix code completion and code linting issues.
PLATFORMIO_DEVICE_LIST - List connected devices.
```

I use it on two big projects and it works great. I hope you'll add it for everyone.

### DIFF
--- a/platformio/ide/tpls/cmake/.gitignore.tpl
+++ b/platformio/ide/tpls/cmake/.gitignore.tpl
@@ -1,0 +1,2 @@
+.pio
+CMakeListsPrivate.txt
--- a/platformio/ide/tpls/cmake/CMakeLists.txt.tpl
+++ b/platformio/ide/tpls/cmake/CMakeLists.txt.tpl
@@ -1,0 +1,83 @@
+# !!! WARNING !!! AUTO-GENERATED FILE, PLEASE DO NOT MODIFY IT AND USE
+# https://docs.platformio.org/page/projectconf/section_env_build.html#build-flags
+#
+# If you need to override existing CMake configuration or add extra,
+# please create `CMakeListsUser.txt` in the root of project.
+# The `CMakeListsUser.txt` will not be overwritten by PlatformIO.
+
+cmake_minimum_required(VERSION 3.2)
+project({{project_name}})
+
+include(CMakeListsPrivate.txt)
+
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/CMakeListsUser.txt)
+include(CMakeListsUser.txt)
+endif()
+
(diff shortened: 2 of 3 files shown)


ivankravets:
Have you seen `platformio init --ide clion`? You can write a script which generates project for CLIon and then removes not used CLIon related files

xeno010:
Sure, but I'd have to delete the .idea folder every time even if I only call ```cmake PLATFORMIO_REBUILD_PROJECT_INDEX (->pio -f -c clion init --ide clion)```

second problem:
```PLATFORMIO_REBUILD_PROJECT_INDEX (-> pio -f -c clion init --ide clion)``` does not work properly in clion. It overwrites each time the files workspace.xml, watcherTasks.xml and modules.xml in the .idea folder. This means that all my custom file guards, my custom settings etc are deleted and I have to restart clion every time.

I develop in CLion and i use ```pio -f -c cmake init --ide cmake``` instead of ```pio -f -c clion init --ide clion```(for rebuilding index) which only updates CMakeLists.txt and CMakeListsPrivate.txt.

mnrce:
This will also solve #2990, because if `CMAKE_EXPORT_COMPILE_COMMANDS` is set (env or in `CMakeLists.txt`), `cmake` will generate `compile_commands.json` for us.

ivankravets:
We implemented https://github.com/platformio/platformio-core/issues/2990

What is a reason to use cmake wrapper around our CLI? We are on the way dropping support for CMake because of these are overheads.

ivankravets:
Please use compilation database https://docs.platformio.org/en/latest/faq.html#compilation-database-compile-commands-json

PR #3072, opened 2019-09-30
Create qtcreator-generic IDE template

### LINKED ISSUE #3046: Qt Creator integration enhancement request
What kind of issue is this?

- [x] **Feature Request**.

- [x] **PlatformIO Core**.

### Configuration

**Operating system**: All

**PlatformIO Version** (`platformio --version`): 4.0.3

### Additional info

The current integration with Qt Creator works quite well with its ability to output `.pro` files.  However, I would like to see it output it as [Qt Creator generic project](https://doc.qt.io/qtcreator/creator-project-generic.html) files instead, with files like `.config`, `.creator`, `.files`, `.include`, etc.  This will allow it to be used with kit configurations that don't include Qt itself, as `.pro` files are more specific to projects using Qt and requires `qmake` and a full Qt installation.  Plus, `qmake` is a dying application which they are gradually replacing with `cmake`.

All of the content of the `.pro` file that PlatformIO creates fits nicely in these generic project files, which Qt Creator will load together as a project.  Just put the predefined macros (or defines) in the `.config` file.  Put the list of project source files in the `.files` file.  And list the include paths in the `.include` file.  And create a `.creator` file with a single `[General]` line.  And if there are build flags, `.cflags` and `.cxxflags` files can be used.

I think this would be a relatively easy change to PlatformIO, but would be a nice improvement in its export to Qt Creator and will allow people to use it with just Qt Creator installed without installing all of Qt too.

dewhisna:
Add qtcreator-generic IDE template.  Fixes #3046 

Since Qt Creator can default to running `make` for generic targets, this template also adds a `Makefile` that will allow Qt Creator to automatically build the project without having to manually modify the project settings to launch `platformio`.

This `qtcreator-generic` IDE template diverges from the `qtcreator` template in that it doesn't require a Qt Kit to be setup that contains `qmake`.  It also outputs compiler flags so that detail from custom scripts that alter the compiler flags can be captured by the code model in Qt Creator.  It does, however, lose the `HOMEDIR` isolation of include paths that the `qtcreator` template provides -- this is because I see no good way to do it with the generic project mechanism since it's not parsed by `qmake`.

Run with: `platformio init --ide qtcreator-generic`


valeros:
Hi @dewhisna ! Thanks a lot for the PR and sorry for the late reply. Indeed, it looks like a good alternative to a full-featured QtCreator project. I played with your template a bit and while there is no issues with simple projects, I noticed that code completion doesn't take into account compiler flags specified in the `.cflags`, `cxxflags` files which is critical for some frameworks that have compiler flags with includes (e.g. `-include`, `-imacros`, etc). Even if you export the C/C++ code model you won't see the flags specified in the `.cflags`, `cxxflags` files. Any thoughts?

dewhisna:
@valeros -- I know it certainly uses all of the `.cflags` and `.cxxflags` when actually compiling and building things, as I've been using this for some time on my PlatformIO projects with all sorts of additional compiler flags, like setting special defines, etc., and they have always made it successfully into the final build.  But, the code completion and clang language model stuff in the editor, as you have seen, do act a bit weird.

To make it even more strange, if you open a code file in a library instead of the main project files, it seems to process them there and shows things correctly (or at least more correctly) even when the main project code files do not.  I've mostly chalked that up to Qt Creator quirks.  The code that it produces and ultimate calling of the compiler works correctly and gets all of the flags passed to the actual compiler.

I hadn't worried to much about it, because the clang code model under Qt Creator tends to crash fairly regularly on me anyway.  And there also seems to be some differences between Linux and Windows.  For PlatformIO, 99.9+% of the time, I'm running on Linux.  But I do use Qt Creator regularly on Windows and the code completion irregularities and peculiarities seem slightly different on the two environments.

In any case, it's a fact that Qt is doing away with QMake and are moving toward CMake only.  So having projects not rely on QMake is the way to go.

valeros:
> it certainly uses all of the .cflags and .cxxflags when actually compiling and building things, as I've been using this for some time on my PlatformIO projects with all sorts of additional compiler flags, like setting special defines, etc., and they have always made it successfully into the final build. 

Not sure I understand, your project is compiled by PlatformIO as specified in the Makefile:
```
default_target: all

all:
	platformio -f -c qtcreator run
clean:
	platformio -f -c qtcreator run --target clean


```
PlatformIO build system is unaware of the files `.cflags, .cxxflags`, so any additional flags specified in these files won't appear in the final compiler commands. It's my understanding that the flags specified in the `.cflags, .cxxflags` are meant to be used solely by QtCreator. Am I missing something?

dewhisna:
Ah yes, you are right -- sorry, I keep bouncing between build systems and forgetting what is where.

But yes, the `.cflags` and `.cxxflags` are there for the code model, and it is documented that way in [the Qt Documentation for Generic Projects](https://doc.qt.io/qtcreator/creator-project-generic.html).  As I mentioned, it strangely seems to work correctly for files in the library folders (i.e. files in the include paths of the `.include` file that aren't explicitly listed in the `.files` file).

I noticed that because I often am developing library code and do it "in place" within an active project so that I can test and debug it before committing it to the library repo.  I'll have PlatformIO clone the git repo in its `libdeps` folder and edit it there within the same Qt Creator session as the main project, and then just commit the changes back to that repo when they are done.  When I do that, the code model in Creator manages to figure things out correctly for those files.  But strangely, not for files explicitly listed in the `.files` file.  And I have never quite figured that one out.

dewhisna:
@valeros -- OK, after much experimenting, I think I finally know what's going on.  The `.cflags` and `.cxxflags` are both getting sent to the Clang Code Model.  However, those flags are gcc flags and not clang flags.  As such, it isn't understanding all of them and it ends up mostly disabling the code model.

However, I did find a scenario that seems to work, at least on my projects here with the STM32F103 core libraries.  And that was by doing the following:

1. Make sure you are on the latest Qt Creator (currently 4.11.1)
2. Set up a Bare Metal Kit in Creator manually selecting the `GCC-arm-none-eabi-platformio` with no Qt version and pointing it at the `~/.platformio/packages/toolchain-gccarmnoneeabi` compiler.
3. Create the generic creator project with the IDE template from this pull request.
4. Run `platformio run -t compiledb` to create a `compile_commands.json` file and add it to the list in the `.files` file from this template.  Note, I had to update to PlatformIO 4.2.2a1 via `pio upgrade --dev` as the release 4.2.0 kept exiting with errors when building the `compile_commands.json` target, but 4.2.2a1 built it successfully.
5. Enable the `CompilationDatabaseProjectManager` plugin in Creator and restart Creator to activate it.  Note that that plugin is still somewhat experimental in Creator.
6. Import both the `compile_commands.json` and the `.creator` generic project file from this template.  Go into the `Projects` settings and manually select the Bare Metal Kit that was created as the build target and enter `make all` and `make clean` for the custom build and clean commands, respectively, if it didn't find them on its own (for me, for some reason it found the `make all` step but not the `make clean`).

After doing this, the project is now correctly running the clang code model and completions on all of the code files from the project, and it has all of the library and core include paths and such too, and seems to have the compiler flags figured out.

Getting the kit and compiler setup was key as the clang code model tools call it with the options to dump all of the compiler defines, and without the correct compiler, it was calling the system gcc compiler instead which exited with errors from the arm options, causing it to fail and not run the code model.  But once it started calling the correct compiler, it seems to have things figured out.

So... from what I'm seeing from this experiment, this IDE template needs to somehow call the PlatformIO code to generate the Compilation Database along with the rest of the project and add it to the files list for Qt Creator to use.  But, I'm not totally sure how to automate that step in the PlatformIO IDE generation template.

CODeRUS:
adding
```
% for include in filter_includes(includes):
{{include}}
% end
```
to `platformio.includes.tpl` fixes libs autocompletion

CODeRUS:
also it's nice to have following `Makefile`:
```
default_target: all

all:
  platformio -f -c qtcreator run

clean:
  platformio -f -c qtcreator run --target clean

buildfs:
  platformio -f -c qtcreator run --target buildfs

upload:
  platformio -f -c qtcreator run --target upload

uploadfs:
  platformio -f -c qtcreator run --target uploadfs
```

CODeRUS:
btw, `.gitignore.tpl` creates `gitignore` file without dot at the beginning. @ivankravets @valeros

belm0:
What needs to be done to revive this PR?

belm0:
@ivankravets if I rebase this PR and confirm it still works, would you merge it?

ivankravets:
Does this PR work for you if you change file structure?

belm0:
I see, a project path has changed since this PR:

`platformio/ide/tpls` --> `platformio/project/tpls`

I'll make a new PR, unless you're willing to grant me contributor status on platformio-core.

ivankravets:
We don't understand the goal of this PR. Why we can't improve https://github.com/platformio/platformio-core/tree/develop/platformio/project/tpls/qtcreator ?

belm0:
> We don't understand the goal of this PR. Why we can't improve https://github.com/platformio/platformio-core/tree/develop/platformio/project/tpls/qtcreator ?

I suspect the original author was thinking to keep `qtcreator` as it is, generating a Qt `.pro` file.  Then anyone who may be reading old docs, articles, or running scripts using `pio` will still have a good experience.

Other than the directory change, I noticed that quoting in the `config` output file needs to be fixed.

ivankravets:
Closing in favor of https://github.com/platformio/platformio-core/pull/4262

PR #3119, opened 2019-10-14
Replace deprecated in python3: iteritems with items and basestr with str

mmccartn:
Fixes:
```
'dict' object has no attribute 'iteritems'
[FAILURE]
```
and
```
name 'basestring' is not defined
```
when run with Python3.5+ on Windows


ivankravets:
Thanks!

PR #3296, opened 2019-12-02
[WIP] Make ManifestSchema compatible with marshmallow >= 3

StefanBruens:
Instead of suppressing the exception, checking for an error and raising
a PlatformIOException, directly raise the PlatformIOException from
handle_error().

Add a small wrapper for Schema.load() to cover the the return value
difference ((data, error) tuple vs data only).

For more information, see
https://marshmallow.readthedocs.io/en/stable/upgrading.html

### DIFF
--- a/platformio/package/manifest/schema.py
+++ b/platformio/package/manifest/schema.py
@@ -15,13 +15,14 @@
 import requests
 import semantic_version
 from marshmallow import Schema, ValidationError, fields, validate, validates
+from marshmallow import __version_info__ as marshmallow_version
 
 from platformio.package.exception import ManifestValidationError
 from platformio.util import memoized
 
 
 class StrictSchema(Schema):
-    def handle_error(self, error, data):
+    def handle_error(self, error, data, *, _many, **_kwargs):
         # skip broken records
         if self.many:
             error.data = [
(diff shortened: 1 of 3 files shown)


ivankravets:
Could you fix a PR to make it compatible between Python 2 and Python 3?

StefanBruens:
Supporting both Python2 and 3 is actually a little bit more work, but I am on it ...

platformio/package/manifest/schema.py
```diff
@@ -15,13 +15,14 @@
...
 from marshmallow import Schema, ValidationError, fields, validate, validates
+from marshmallow import __version_info__ as marshmallow_version
 
 from platformio.package.exception import ManifestValidationError
 from platformio.util import memoized
 
 
 class StrictSchema(Schema):
-    def handle_error(self, error, data):
+    def handle_error(self, error, data, *, _many, **_kwargs):
```
hmaarrfk: Python 2.7 doesn't have keyword only arguments 

https://www.python.org/dev/peps/pep-3102/
```suggestion
    def handle_error(self, error, data, _many, **_kwargs):
```

platformio/package/manifest/schema.py
```diff
@@ -142,9 +145,17 @@ class ManifestSchema(Schema):
...
+    # Wrapper to encapsulate differences between marshmallow 2 and 3
+    # load() method
+    def load_manifest(self, manifest):
+        if marshmallow_version >= (3, ):
+            data = self.load(manifest)
+        else:
+            data, _ = self.load(manifest)
+        return data
+
+    def handle_error(self, error, data, *, _many, **_kwargs):
```
hmaarrfk:
```suggestion
    def handle_error(self, error, data, _many, **_kwargs):
```

No keyword only arguments in Python 2.7

hmaarrfk:
Sorry for jumping in, just did a quick browse and noticed at least one feature that isn't available in Python 2.7.

PR #3305, opened 2019-12-14
Include LICENSE file in distribution

hmaarrfk:


ivankravets:
We use low-level `setup.py` configuration for setuptools to make it compatbile with old Python distirbution. Could you use it too?

hmaarrfk:
How old of a version are you targeting? this should be compatible with 2.7

hmaarrfk:
Superseded by https://github.com/platformio/platformio-core/pull/3325

Issue #3304, opened 2019-12-13
Randomly rebuilding mbed library


HeathPetersen:
The mbed library is being rebuilt at random times WITHOUT ANY file changes.  It is easily reproducible as follows:

```
mkdir x; cd x
pio init --board nucleo_l432kc --project-option "framework=mbed"
cat >src/main.cpp <<END
#include <mbed.h>
int main() {}
END
pio run
```

After that, consecutive `pio run` commands may or may not recompile the mbed libraries.

There appear to be multiple mentions of this bug from other people.  For examples, see the following:
* [Always full rebuild allegedly after V4.0.0 - is my platformio.ini file deficient?](https://community.platformio.org/t/always-full-rebuild-allegedly-after-v4-0-0-is-my-platformio-ini-file-deficient/9327)
* [PIO/VSCode recompiles everything, every time](https://community.platformio.org/t/pio-vscode-recompiles-everything-every-time/10135)
* [Entire MBED recompiles every time](https://community.platformio.org/t/entire-mbed-recompiles-every-time/6766)
* [Randomly rebuilding all mbed libraries](https://community.platformio.org/t/randomly-rebuilding-all-mbed-libraries/9628/3)

Now, I admit that some of these may be something else.  However, the fact that multiple people are mentioning this, it only seems to be mentioned related to the mbed framework, and it is easily reproduced as shown above leads me to believe there is a genuine bug.

I am running platformio on Ubuntu 19.04 - fully upgraded.  The `platformio` / `pio` commands start with `#!/usr/bin/python3` so it should be using python 3.  `python3 --version` returns `Python 3.7.3`.  I have tested using platformio v4.1.0 and v4.1.1b5.

FYI - I did a very lengthy / detailed comparison of the whole project and .platformio directorys' timestamps and file sizes to see if something was changing without my knowledge.  I was unable to find anything changing outside of the generated files during runs.

nmathon:
Hi, I published a topic here on the same subject :
https://community.platformio.org/t/build-compiles-all-source-files-every-time/11262

And I found this anomaly:
https://gavv.github.io/articles/sconsign-bug/

I tried to modify the line 144 of the file .platformio/penv/lib/python3.7/site-packages/platformio/builder/main.py
And I replaced:
`
env.SConsignFile(
    join("$BUILD_DIR", ".sconsign.py%d%d" % (sys.version_info[0], sys.version_info[1]))
)
`
by
`
env.SConsignFile(
[...]

PR #3324, opened 2019-12-30
Project is always compiling all the files

nmathon:
This is related to this bug : https://gavv.github.io/articles/sconsign-bug/
If the project contains many files, like a project based on the mbedos project, it will always compile every files and debugging is not possible anymore due to this error:
`
*** [.pio/build/disco_f746ng/FrameworkMbed/targets/TARGET_STM/TARGET_STM32F7/device/stm32f7xx_hal_cryp_ex.o] /Users/username/src/projectname/gsboard-v2/.pio/build/disco_f746ng/FrameworkMbed/targets/TARGET_STM/TARGET_STM32F7/device/stm32f7xx_hal_cryp_ex.o: No such file or directory
FileNotFoundError: [Errno 2] No such file or directory: '/Users/username/src/projectname/gsboard-v2/.pio/build/disco_f746ng/.sconsign.py37.dblite':
  File "/Users/username/.platformio/packages/tool-scons/script/../engine/SCons/Script/Main.py", line 1374:
    _exec_main(parser, values)
`

This small change correct this problem

Related issue:
https://github.com/platformio/platformio-core/issues/3304

### DIFF
--- a/platformio/builder/main.py
+++ b/platformio/builder/main.py
@@ -142,7 +142,7 @@ env.LoadPioPlatform()
 
 env.SConscriptChdir(0)
 env.SConsignFile(
-    join("$BUILD_DIR", ".sconsign.py%d%d" % (sys.version_info[0], sys.version_info[1]))
+    join("$BUILD_DIR", ".sconsign.dblite")
 )
 
 for item in env.GetExtraScripts("pre"):


ivankravets:
@valeros could you check mbed builder?

ivankravets:
We rebuild project in debug mode for PIO Unified Debugger. You can avoid rebuilding with `build_type = debug`. See https://docs.platformio.org/en/latest/projectconf/build_configurations.html

valeros:
Hi @nmathon ! Could you please upload somewhere a minimal project to reproduce the issue?

nmathon:
Hi ! This issue is not reproductible with a minimal project. As written here, it occurs when there is many files in the project, and the mbed framework contains already a lot, so if you have 100 additional source files (like my project), you will have this bug : https://gavv.github.io/articles/sconsign-bug/

ivankravets:
Let's move the discussion here https://github.com/platformio/platform-ststm32/issues/331

PR #3338, opened 2020-01-13
pio-test: pass --verbose to the run command context

mcspr:
Makes `pio test --verbose` output the same as `pio run --verbose`, showing exact shell commands that are executed.
<s>Also, `--silent` flag probably needs to be exposed too.</s>

---

In case of `native` test, what I now see instead of:
```
Linking .pio/build/test/program
```
Is this:
```
g++ -o .pio/build/test/program .pio/build/test/test/output_export.o .pio/build/test/test/my_test.o -L.pio/build/test -Wl,--start-group .pio/build/test/libUnityTestLib.a -Wl,--end-group
```


mcspr:
On a second though, I think my proposal is to make `test --verbose` == `run --verbose`, instead of a simple `run`. Last commit restores old behavior with silent tests

ivankravets:
Thanks, @mcspr!

PR #3415, opened 2020-03-13
Ensure all commands in compilation_commands.json use absolute paths.

mincrmatt12:
By placing the `where_is_program` call into this function, all references to the compiler will be made absolute, instead of just ones in the top environment. For example, previously all references to the compiler for user source code would not use the full path in the compilation database, which broke `clangd`'s detection of system includes.

### DIFF
--- a/platformio/builder/tools/compilation_db.py
+++ b/platformio/builder/tools/compilation_db.py
@@ -75,6 +75,16 @@ def makeEmitCompilationDbEntry(comstr):
         :param env: Environment for use building this node
         :return: target(s), source(s)
         """
+        
+        # Resolve absolute path of toolchain
+        for cmd in ("CC", "CXX", "AS"):
+            if cmd not in env:
+                continue
+            if os.path.isabs(env[cmd]):
+                continue
+            env[cmd] = where_is_program(
+                env.subst("$%s" % cmd), env.subst("${ENV['PATH']}")
+            )
 
         dbtarget = __CompilationDbNode(source)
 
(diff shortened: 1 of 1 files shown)


ivankravets:
Could you provide a simple project or steps how to reproduce this issue?

mincrmatt12:
Sure. To recreate the issue:

1. Create a new project with any IDE and (for example) the `megaatmega2560` board (although any board/framework I've tested has had similar results)
2. Create an empty `main.cpp` file in the `src` folder.
3. Create a compilation database with `pio run -t compiledb`

The generated file will end with an entry for `src/main.cpp`, and will incorrectly set the command to something like 
```"command": "avr-g++ -o .pio/build/megaatmega2560/src/main.cpp.o -c -fno-exceptions -fno-threadsafe-statics <...other stuff...> src/main.cpp"```, failing to expand the full path of `avr-g++`.

The rest of the entries (for library code / framework code) correctly have entries like
```"command": "/root/.platformio/packages/toolchain-atmelavr/bin/avr-g++ -o .pio/build/megaatmega2560/FrameworkArduino/new.cpp.o -c -fno-exceptions -fno-threadsafe-statics <...other stuff...> /root/.platformio/packages/framework-arduino-avr/cores/arduino/new.cpp"```

(the strange paths are because I did this in a docker container, but the issue still occurs outside of one)

ivankravets:
Thank you so much!

ivankravets:
What is your use case for `compilation_commands.json`?

mincrmatt12:
I use it for completion with vim and clangd (ccls didn't work for other projects)

ivankravets:
Does it work well?

mincrmatt12:
With this patch, and the appropriate command line flags (`-query-driver=/path/to/g++/in/.platformio/folder`, otherwise it doesn't find the standard library headers) it works great.

Issue #3435, opened 2020-03-22
Fish shell completions


sdaitzman:
### Configuration

**Operating system**: macOS

**PlatformIO Version** (`platformio --version`): PlatformIO, version 4.0.0

### Description of problem
I would love to have Fish Shell completions!

ivankravets:
Have you tried this https://github.com/click-contrib/click-completion? if it works, we will add to our docs.

ivankravets:
Please re-test with `pio upgrade --dev`. See docs how to use it https://docs.platformio.org/en/latest/core/userguide/misc/completion/index.html

mcspr:
Thanks! re #3449, running `pio misc completion install` multiple times should overwrite instead of append to the completion file? `append` kwarg seems to do it.
```
$ cat ~/.config/fish/completions/pio.fish
complete --command pio --arguments "(env _PIO_COMPLETE=complete-fish COMMANDLINE=(commandline -cp) pio)" -f
complete --command pio --arguments "(env _PIO_COMPLETE=complete-fish COMMANDLINE=(commandline -cp) pio)" -f
complete --command pio --arguments "(env _PIO_COMPLETE=complete-fish COMMANDLINE=(commandline -cp) pio)" -f
```

New versions works for me in fish, as well as raw click-complete patch from the referenced PR. Sadly, dynamic things are still missing, like `--board` or `--environment` lists. I have not explored that part more in-depth.

---

There is also a weird bug when typing `pio foo .` and TAB
```python3
  File "/home/builder/runtime/platformio-core/platformio/commands/__init__.py", line 64, in get_command
    mod = __import__(mod_path, None, None, ["cli"])
ModuleNotFoundError: No module named 'platformio.commands.foo'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/builder/runtime/platformio-core/platformio/__main__.py", line 109, in main
    cli()  # pylint: disable=no-value-for-parameter
  File "/home/builder/.platformio/penv/lib64/python3.8/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/builder/.platformio/penv/lib64/python3.8/site-packages/click/core.py", line 712, in main
    _bashcomplete(self, prog_name, complete_var)
  File "/home/builder/.platformio/penv/lib64/python3.8/site-packages/click_completion/patch.py", line 135, in _shellcomplete
    do_fish_complete(cli, prog_name)
[...]

PR #3449, opened 2020-04-06
(wip) Shell complete

mcspr:
As suggested in #3435, proof of concept that click-complete works by c/p example code from click-completion
I have tested fish, seems to work so far. Only one strange issue so far - it always expands to a full option in cases like `pio run -e` -> `pio run --environment`

*edit:* see `pio completion show` & `pio completion install` for the shell code that it generates

### DIFF
--- a/platformio/commands/__init__.py
+++ b/platformio/commands/__init__.py
@@ -15,6 +15,8 @@
 import os
 
 import click
+import click_completion
+click_completion.init()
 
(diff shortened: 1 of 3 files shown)


ivankravets:
Resolved in https://github.com/platformio/platformio-core/issues/3435#issuecomment-626347972

Please re-test.

PR #3497, opened 2020-05-08
Save summary data to local session.

ShahRustam:


platformio/commands/account/client.py
```diff
@@ -54,6 +54,15 @@ def get_refresh_token():
...
 
+    @staticmethod
+    def delete_data_from_local_session(key):
+        try:
+            account = app.get_state_item("account")
+            del account[key]
+            app.set_state_item("account", account)
+        except KeyError:
+            pass
+
```
ivankravets, line 65: rename to `delete_local_state`

PR #3537, opened 2020-06-01
Fix #3535 - Spurious rebuilds

### LINKED ISSUE #3535: Spurious rebuilds - partial fix
- [X ] **PlatformIO Core**.

### Configuration

**Windows 10**:

**PlatformIO, version 4.3.4**

### Description of problem

I am using platformio to build the source code in https://github.com/bdring/Grbl_Esp32

Even with no changes to the source tree, it often recompiles most or all of the .cpp files and relinks the .elf file.

I can be reproduced by executing "pio run" several times in succession.  It should compile everything the first time, then subsequent times it should skip compilation and linking, since nothing has changed.  The problem is somewhat random.  There is a small chance that it will work correctly, but usually it performs unnecessary recompiles.

Using the "--debug=explain" flag in SCONSFLAGS, I determined that the root cause of the problem is that the order of "-I <include dir>" flags on the cpp command line changes from build to build.  It is always the same list of include directories, but the order is different, triggering SCons to recompile.  This is especially troublesome in conjunction with the "longcmd" trick that is used to work around the Windows command line limitation.

I fixed the spurious recompile problem by sorting the list of include files, adding the line marked with "+" in piomaxlen.py 

```
def long_incflags_hook(env, incflags):
    _incflags = env.subst(incflags).replace("\\", "/")
+    _incflags = " -I".join(sorted(_incflags.split(" -I")))
```

For the related spurious-relinking problem, I determined that the root cause is similar - the order of the library file (*.a) on the linker command line changes from build to build.  I have tried to find a similar fix by sorting the list of libraries, but I have not found the right place in the code to do that.

MitchBradley:
This change eliminates spurious recompiles and relinks caused by indeterminate order of includes on the cpp command and indeterminate order of libraries on the linker command line.

I tried to follow the contribution guidelines, but the "make test" step failed even without my changes, so I do not trust the tests.

### DIFF
--- a/platformio/builder/tools/piolib.py
+++ b/platformio/builder/tools/piolib.py
@@ -1006,6 +1006,7 @@ def ConfigureProjectLibBuilder(env):
 
     def _print_deps_tree(root, level=0):
         margin = "|   " * (level)
+        root.depbuilders.sort(key=lambda lb: lb.path)
         for lb in root.depbuilders:
             title = "<%s>" % lb.name
             vcs_info = _get_vcs_info(lb)
--- a/platformio/builder/tools/piomaxlen.py
+++ b/platformio/builder/tools/piomaxlen.py
@@ -43,6 +43,7 @@ def long_sources_hook(env, sources):
 
 def long_incflags_hook(env, incflags):
     _incflags = env.subst(incflags).replace("\\", "/")
+    _incflags = " -I".join(sorted(_incflags.split(" -I")))
     if len(_incflags) < MAX_LINE_LENGTH:
         return incflags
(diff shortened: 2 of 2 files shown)


ivankravets:
Thank you so much for your PR! What is your `platformio.ini`? It seems like there is an issue with LDF. We need to find a root of problem.

MitchBradley:
The platformio.ini can be found in the github repo mentioned in the first message of the referenced issue

ivankravets:
Resolved in https://github.com/platformio/platformio-core/issues/3535

PR #3576, opened 2020-06-30
Add .ccls to .gitignore (vim and emacs)

pockata:
Adds `.ccls` in `.gitignore` for Vim (introduced in 223a85baca14150400e324ae39196be87822e265) and Emacs (introduced in 7a07a2e63eb51f8e403ecaf0fb2e4029d7b54744).

The file `.ccls` doesn't need to be tracked in git as it contains user/machine-related configurations (similar to `.clang_complete`).


ivankravets:
Thank you so much!

PR #3606, opened 2020-07-21
Check lower case http_proxy environment variables

Sh3Rm4n:
Not all systems have their proxy environment variable set as uppercase. According to [stackoverflow](https://unix.stackexchange.com/questions/212894/whats-the-right-format-for-the-http-proxy-environment-variable-caps-or-no-ca) most of them don't. This is also the case for the system, I'm working with currently.

### DIFF
--- a/platformio/commands/home/helpers.py
+++ b/platformio/commands/home/helpers.py
@@ -59,8 +59,10 @@ def is_twitter_blocked():
     ip = "104.244.42.1"
     timeout = 2
     try:
-        if os.getenv("HTTP_PROXY", os.getenv("HTTPS_PROXY")):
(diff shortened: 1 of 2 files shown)


ivankravets:
Thanks!

I removed unused code and implemented a little bit different. Could you re-test with `pio upgrade --dev`?

Sh3Rm4n:
Thanks. Your implementation is much simpler than mine :sweat_smile: 
I'll test tomorrow, when I've access to the machine again.

Sh3Rm4n:
Working like expected 👍

ivankravets:
Great! Thanks for the report!

Happy coding with PlatformIO!

PR #3615, opened 2020-08-02
Support for switching between multiple environments in VSCode

drzony:
This commit changes the way c_cpp_properties.json is generated, creating a separate build
configuration for each environment. This allows to select an environment using VSCode "build
configurations" option


drzony:
This is my initial proposition, however it changes the default behavior of the `pio init` command, so some changes may be required.
Original: When `pio init` is used without `-e` switch, the `best` config is selected
Current: When `pio init` is used without `-e` switch, all envs in `default_envs` are created

ivankravets:
Thank you so much for your work! There is PlatformIO IDE v2.0-beta. We wait for PIO Core 4.4 release and will release 2.0 version. Please test it https://community.platformio.org/t/platformio-ide-for-vscode-2-0-beta-help-us-with-testing/14097

drzony:
@ivankravets The new project switcher does not handle IntelliSense like my change did. The problem is that c_cpp_properties.json contains includes for the first platform only. I'm using ESP8266 and ESP32 in one project, my solution allowed me to use vscode configuration switcher to switch between platforms and IntelliSense and other features handled it properly, with 2.0 beta I always get includes/defines for the first platform in platformio.ini. Are there any other setting that need to be enabled for it to work?

ivankravets:
Could you try the latest Beta 6? 

- https://community.platformio.org/t/platformio-ide-for-vscode-2-0-beta-help-us-with-testing/14097/36?u=ivankravets

drzony:
I've just installed it and the result is the same. c_cpp_properties.json still contains esp32 includes (I switched to esp8266) and the intellisense still works as if using esp32

ivankravets:
What is your OS?

drzony:
Windows

ivankravets:
This is our bug with multi-config parsing. We will fix it. As a temporary solution, you can use one `platformio.ini`. Thanks for the report!

ivankravets:
Please open PlatformIO CLI (see terminal icon on the blue bottom toolbar) and type
```
pio upgrade --dev
```

Restart VSCode.

Does it work now?

drzony:
Yes, now it works.

ivankravets:
Happy coding with PlatformIO! 😊🎊

PR #3642, opened 2020-08-28
Fix link to FAQ sections

dirkmueller:
Use consistently the same host and url and fix one unmatched anchor.


ivankravets:
Thanks for the PR!

PR #3650, opened 2020-09-03
Exclude tests from python package setup

Jakeler:
The addition of the `tests/__init__.py` makes `find_packages()` recognize `tests` as additional python package. Now this unexpected module gets also installed by pip, possibly into system site-packages. This could cause conflicts with other packages, so I suggest excluding it again.


ivankravets:
Thanks for the PR!

PR #3692, opened 2020-09-29
fix quoting of defines in ccls template

micampe:
This change fixes quoted strings in defines in the ccls vim template.

```diff
--DARDUINO_BOARD=&quot;PLATFORMIO_D1_MINI&quot;
+-DARDUINO_BOARD="PLATFORMIO_D1_MINI"
```


ivankravets:
Thanks for the PR!

PR #3753, opened 2020-12-04
Verify the Returncode of Checks

shaardie:
This patch introduces the verification of returncodes of checks.
Despite the fact, that the called tool itself could have a returncode
unequal 0 and the check still works as expected, there are some
returncodes like 126 and 127, which indicate that the binary was not
properly executed. So this patch checks for those returncodes.

This is the case under, e.g. Debian (Docker container) when running `clang-tidy`:

```bash
# Setup
$ pip install platformio
$ mkdir test
$ cd test
$ pio init -e native -O "check_tool=clangtidy" -O "platform=native"
# Execute check
$ pio check -e native
Checking native > clangtidy (platform: native)
-----------------------------------------------------------
No defects found
=============== [PASSED] Took 0.00 seconds ===============

Environment    Tool       Status    Duration
-------------  ---------  --------  ------------
native         clangtidy  PASSED    00:00:00.004
=============== 1 succeeded in 00:00:00.004 ===============
# But the binary is not executable, see
$ ~/.platformio/packages/tool-clangtidy/clang-tidy 
/root/.platformio/packages/tool-clangtidy/clang-tidy: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
$ echo $?
127
```

### DIFF
--- a/platformio/commands/check/tools/base.py
+++ b/platformio/commands/check/tools/base.py
@@ -200,12 +200,22 @@ class CheckToolBase(object):  # pylint: disable=too-many-instance-attributes
             if self.options.get("verbose"):
                 click.echo(" ".join(cmd))
 
-            proc.exec_command(
+            result = proc.exec_command(
                 cmd,
                 stdout=proc.LineBufferedAsyncPipe(self.on_tool_output),
                 stderr=proc.LineBufferedAsyncPipe(self.on_tool_output),
             )
-
+            # Check if at least the binary was executed properly.
+            # 126 - A file to be executed was found,
+            #       but it was not an executable utility.
+            # 127 - A utility to be executed was not found.
+            if result["returncode"] in (126, 127):
+                if self.options.get("verbose"):
+                    click.echo(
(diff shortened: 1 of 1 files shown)


shaardie:
Seems like the pipeline was already broken in the `develop` branch.

valeros:
Hi @shaardie ! Thanks for the PR and pointing out the issue. Indeed, there might be cases when we are not be able to perform check operation due to problems with binaries. IMO, checking only the special return codes like 126 and 127 doesn't seem cross-platform enough, for example on Windows it might crash with bizarre memory access errors like `0xC0000374` when a dynamic library is missing. Instead, I propose the opposite, only acceptable return values should not lead to check failure (each tool can individually configure acceptable conditions). 

I pushed some changes in https://github.com/platformio/platformio-core/commit/6e958b8415f850b23fb5404b54ec1af59e6a56bd and it would be great if you could test the latest `platformio` from the dev branch in your Docker container.

PR #3756, opened 2020-12-05
Fix silent hang when custom `debug_server` is not found

nchevsky:
With `debug_tool = custom` and `debug_server` set to an invalid path,
`pio debug` would hang indefinitely, not surface any errors, and ignore
SIGINT (i.e. it couldn't even be killed with Ctrl+C).

This change attaches an errback to the `GDBClient::spawn()` call in
`platformio/commands/debug/command.py::cli()`. The errback catches exception
`DebugInvalidOptionsError` asynchronously raised by `DebugServer::spawn()`
when `server_executable` cannot be found. This exception was being suppressed
because (a) there was no errback and (b) as a result of being raised from
a Deferred, the exception couldn't bubble up into the main `try...except`
handler in `platformio/__main__.py`.

The new errback stores the asynchronously caught exception and stops the
reactor. When the `reactor.run()` call in `cli()` then returns, we retrieve
the stored exception and raise it synchronously, allowing it to be caught by
the main `try...except` handler in `platformio/__main__.py` which gracefully
prints the error message and terminates the script with an error exit code.

### DIFF
--- a/platformio/commands/debug/command.py
+++ b/platformio/commands/debug/command.py
@@ -166,10 +166,22 @@ def cli(ctx, project_dir, project_conf, environment, verbose, interface, __unpro
     # pylint: disable=import-outside-toplevel
     from platformio.commands.debug.process.client import GDBClient, reactor
 
-    client = GDBClient(project_dir, __unprocessed, debug_options, env_options)
-    client.spawn(ide_data["gdb_path"], ide_data["prog_path"])
+    # pylint: disable=no-member
+    GDBClient(project_dir, __unprocessed, debug_options, env_options) \
+    .spawn(ide_data["gdb_path"], ide_data["prog_path"]) \
+    .addErrback(handleFailure)
 
     signal.signal(signal.SIGINT, lambda *args, **kwargs: None)
     reactor.run()
 
+    if hasattr(handleFailure, "exception"):
+        raise handleFailure.exception
+
     return True
+
+def handleFailure(failure):
+    handleFailure.exception = failure.value
+
+    # pylint: disable=import-outside-toplevel
+    from platformio.commands.debug.process.client import reactor
+    reactor.callWhenRunning(reactor.stop)


platformio/commands/debug/command.py
```diff
@@ -166,10 +166,22 @@ def cli(ctx, project_dir, project_conf, environment, verbose, interface, __unpro
     # pylint: disable=import-outside-toplevel
     from platformio.commands.debug.process.client import GDBClient, reactor
 
-    client = GDBClient(project_dir, __unprocessed, debug_options, env_options)
-    client.spawn(ide_data["gdb_path"], ide_data["prog_path"])
+    # pylint: disable=no-member
+    GDBClient(project_dir, __unprocessed, debug_options, env_options) \
+    .spawn(ide_data["gdb_path"], ide_data["prog_path"]) \
+    .addErrback(handleFailure)
```
nchevsky, line 172: This new errback catches exception `DebugInvalidOptionsError` asynchronously raised by `DebugServer::spawn()`
when `server_executable` cannot be found. The exception was being suppressed because (a) there was no errback and (b) as a result of being raised from a Deferred, the exception couldn't bubble up into the main `try...except` handler in `platformio/__main__.py`.

platformio/commands/debug/command.py
```diff
@@ -166,10 +166,22 @@ def cli(ctx, project_dir, project_conf, environment, verbose, interface, __unpro
...
+        raise handleFailure.exception
+
     return True
+
+def handleFailure(failure):
+    handleFailure.exception = failure.value
+
+    # pylint: disable=import-outside-toplevel
+    from platformio.commands.debug.process.client import reactor
+    reactor.callWhenRunning(reactor.stop)
```
nchevsky, line 187: Stopping the reactor fixes the hang, allowing `cli()` to terminate and return control to `platformio/__main__.py` which will now gracefully print the exception and exit with the appropriate error code.

platformio/commands/debug/command.py
```diff
@@ -166,10 +166,22 @@ def cli(ctx, project_dir, project_conf, environment, verbose, interface, __unpro
...
     signal.signal(signal.SIGINT, lambda *args, **kwargs: None)
     reactor.run()
 
+    if handleFailure.exception:
+        raise handleFailure.exception
+
     return True
+
+def handleFailure(failure):
+    handleFailure.exception = failure.value
```
nchevsky, line 183: We store the asynchronous exception so that it can be retrieved synchronously in `cli()` once the call to `reactor.run()` returns (after the reactor is stopped by `handleFailure()`).

platformio/commands/debug/command.py
```diff
@@ -166,10 +166,22 @@ def cli(ctx, project_dir, project_conf, environment, verbose, interface, __unpro
...
+    # pylint: disable=no-member
+    GDBClient(project_dir, __unprocessed, debug_options, env_options) \
+    .spawn(ide_data["gdb_path"], ide_data["prog_path"]) \
+    .addErrback(handleFailure)
 
     signal.signal(signal.SIGINT, lambda *args, **kwargs: None)
     reactor.run()
 
+    if handleFailure.exception:
+        raise handleFailure.exception
```
nchevsky, line 178: By synchronously raising the asynchronously caught exception here, we allow it to be handled by the main `try...except` block in `platformio/__main__.py` which will now gracefully print the exception and exit with the appropriate error code.

ivankravets:
Thanks for the PRs! We made huge refactoring of PlatformIO Unified Debugger. See https://github.com/platformio/platformio-core/blob/develop/HISTORY.rst

Please re-retest the latest development version via `pio upgrade --dev`.

PR #3757, opened 2020-12-05
Create `_gdbsrc_dir` temporary directory only if debug server launches successfully

nchevsky:
This change postpones the creation of the `_gdbsrc_dir` temporary directory until **after** the debug server has successfully launched.

Previously, `_gdbsrc_dir` would be unconditionally created as soon as the `GDBClient` class was instantiated. If anything then went wrong in launching the debug server (e.g. due to an invalid user-provided `debug_server` with `debug_tool = custom`), the temporary directory would be left behind and never cleaned up.

### DIFF
--- a/platformio/commands/debug/process/client.py
+++ b/platformio/commands/debug/process/client.py
@@ -54,7 +54,7 @@ class GDBClient(BaseProcess):  # pylint: disable=too-many-instance-attributes
 
         if not isdir(get_project_cache_dir()):
             os.makedirs(get_project_cache_dir())
-        self._gdbsrc_dir = mkdtemp(dir=get_project_cache_dir(), prefix=".piodebug-")
+        self._gdbsrc_dir = None
 
         self._target_is_run = False
         self._auto_continue_timer = None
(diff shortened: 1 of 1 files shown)


ivankravets:
Thank you so much for the PR!

>  If anything then went wrong in launching the debug server (e.g. due to an invalid user-provided debug_server with debug_tool = custom), the temporary directory would be left behind and never cleaned up.

Should not be this folder deleted in this place?

https://github.com/platformio/platformio-core/blob/develop/platformio/commands/debug/process/client.py#L185

nchevsky:
> > If anything then went wrong in launching the debug server (e.g. due to an invalid user-provided debug_server with debug_tool = custom), the temporary directory would be left behind and never cleaned up.
> 
> Should not be this folder deleted in this place?
> 
> https://github.com/platformio/platformio-core/blob/develop/platformio/commands/debug/process/client.py#L185

When the debug server fails to launch and an exception is raised [here](https://github.com/platformio/platformio-core/blob/d485703768a5a58bfa2240082d175f31c23761e9/platformio/commands/debug/process/client.py#L80),  `processEnded()` doesn't execute (because the process never started in the first place) and the clean-up code isn't run.

PR #3779, opened 2020-12-29
Add compileCommands to `c_cpp_properties.json`

mickael9:
This makes vscode aware of the location of the `compile_commands.json` file.

### DIFF
--- a/platformio/builder/tools/pioide.py
+++ b/platformio/builder/tools/pioide.py
@@ -167,6 +167,7 @@ def DumpIDEData(env, globalenv):
         "gdb_path": where_is_program(env.subst("$GDB"), env.subst("${ENV['PATH']}")),
         "prog_path": env.subst("$PROG_PATH"),
         "svd_path": _get_svd_path(env),
+        "compile_commands_path": env.subst("$COMPILATIONDB_PATH"),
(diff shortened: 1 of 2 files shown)


ivankravets:
Thanks for the PR!

Do you have any issues without this PR? Just curious. I’ve never heard about this option for VSCode C/C++ extension.

mickael9:
The intention is to allow better discovery of framework (espidf in my case) sources (when doing F12 on a framework function it will go to the implementation instead of the declaration).

However, you might not want to merge this just yet, because I just discovered that the `compile_commands.json` generated by platformio does not actually match the compilation commands that are executed on build (the flags specified in build_flags are not there). I don't know if this is the case for other platforms than esp32 but this could actually degrade the user experience

mickael9:
It looks like I got confused, the strange `compile_commands.json` I was observing was generated by the esp-idf framework during CMake configuration (not by PlatformIO). The one generated by PlatformIO has the right compiler commands, but it isn't built automatically so by default you get a wrong one.

ivankravets:
Yes, exactly. We provide enough information for C/C++ extension.

Happy coding with PlatformIO in 2021 🎅

PR #3786, opened 2021-01-03
Add udev rule for Atmel AVR Dragon

sephalon:
The Atmel AVR Dragon ist not sold anymore, but used to be a much cheaper alternative to a JTAGICE; therefore it's still in use by some people (like me ;-)


ivankravets:
Thanks for the PR!

Issue #3770, opened 2020-12-21
pio update command uninstalls GCC needed by pio run


cbc02009:
### Configuration

**Operating system**: Windows

**PlatformIO Version** (`platformio --version`): PlatformIO, version 5.0.3

### Description of problem
Every time I run the `pio update` command it uninstalls GCC, so the next time I run `pio run` it has to reinstall GCC

### If problems with PlatformIO Build System:

**The content of `platformio.ini`:**
```[env:adafruit_feather_nrf52840]
platform = nordicnrf52
board = adafruit_feather_nrf52840
framework = zephyr
upload_protocol = jlink
```

### Additional info

console output of `pio update`:
```Updating platformio/contrib-piohome           3.3.1 @ ~3.3.1                     [Up-to-date]
Updating platformio/contrib-pysite            2.39.201019 @ ~2.39.0              [Up-to-date]
Updating platformio/tool-unity                1.20500.200612 @ ~1.20500.0        [Up-to-date]
Updating platformio/tool-scons                4.40001.0 @ ~4.40001.0             [Up-to-date]

Platform Manager
================
Platform atmelsam
--------
Updating platformio/atmelsam                  5.0.2                              [Up-to-date]
Updating platformio/framework-zephyr          2.20400.201210 @ ~2.20400.0        [Up-to-date]
Updating platformio/framework-zephyr-cmsis    0.0.0-alpha+sha.542b2296e6 @ 0.0.0-alpha+sha.542b2296e6[Up-to-date]
Updating platformio/framework-zephyr-canopennode 0.0.0-alpha+sha.468d350028 @ 0.0.0-alpha+sha.468d350028[Up-to-date]
Updating platformio/framework-zephyr-civetweb 0.0.0-alpha+sha.99129c5efc @ 0.0.0-alpha+sha.99129c5efc[Up-to-date]
Updating platformio/framework-zephyr-fatfs    0.0.0-alpha+sha.13697783bf @ 0.0.0-alpha+sha.13697783bf[Up-to-date]
Updating platformio/framework-zephyr-hal-atmel 0.0.0-alpha+sha.9f0a699a2f @ 0.0.0-alpha+sha.9f0a699a2f[Up-to-date]
Updating platformio/framework-zephyr-hal-st   0.0.0-alpha+sha.5b3ec3e182 @ 0.0.0-alpha+sha.5b3ec3e182[Up-to-date]
[...]

PR #3809, opened 2021-01-23
pm-update: respect only_check value

mcspr:
Related to the https://github.com/platformio/platformio-core/issues/3770
Don't uninstall the package when user-provided --dry-run / --only-check param (i.e. only_check var is True)

### DIFF
--- a/platformio/package/manager/platform.py
+++ b/platformio/package/manager/platform.py
@@ -118,16 +118,17 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
             )
 
         p.update_packages(only_check)
-        self.cleanup_packages(list(p.packages), only_check)
+        if not only_check:
+            self.cleanup_packages(list(p.packages))
 
-        if missed_pkgs:
-            p.install_packages(
-                with_packages=list(missed_pkgs), skip_default_package=True
-            )
+            if missed_pkgs:
+                p.install_packages(
+                    with_packages=list(missed_pkgs), skip_default_package=True
+                )
 
         return new_pkg or pkg
 
-    def cleanup_packages(self, names, only_check):
+    def cleanup_packages(self, names):
         self.memcache_reset()
         deppkgs = {}
         for platform in PlatformPackageManager().get_installed():
@@ -141,7 +142,7 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
         for pkg in pm.get_installed():
             if pkg.metadata.name not in names:
                 continue
-            if not only_check and (
+            if (
                 pkg.metadata.name not in deppkgs
                 or pkg.metadata.version not in deppkgs[pkg.metadata.name]
             ):
--- a/platformio/package/manager/platform.py
+++ b/platformio/package/manager/platform.py
@@ -118,7 +118,7 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
             )
 
         p.update_packages(only_check)
-        self.cleanup_packages(list(p.packages))
+        self.cleanup_packages(list(p.packages), only_check)
 
         if missed_pkgs:
             p.install_packages(
@@ -127,7 +127,7 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
 
         return new_pkg or pkg
 
-    def cleanup_packages(self, names):
+    def cleanup_packages(self, names, only_check):
         self.memcache_reset()
         deppkgs = {}
         for platform in PlatformPackageManager().get_installed():
@@ -141,7 +141,7 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
         for pkg in pm.get_installed():
             if pkg.metadata.name not in names:
                 continue
-            if (
+            if not only_check and (
                 pkg.metadata.name not in deppkgs
                 or pkg.metadata.version not in deppkgs[pkg.metadata.name]
             ):
--- a/platformio/package/manager/platform.py
+++ b/platformio/package/manager/platform.py
@@ -146,6 +146,8 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
                 pkg.metadata.name not in deppkgs
                 or pkg.metadata.version not in deppkgs[pkg.metadata.name]
             ):
+                if pkg.metadata.spec.url or pkg.metadata.spec.owner != "platformio":
+                    continue
                 try:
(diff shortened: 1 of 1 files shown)


mcspr:
Oh. There's another invocation, but without the CLI params access. Don't really know what to do then :/
https://github.com/platformio/platformio-core/blob/65e67b64bd8d2e06e7fd9b922bb815789f46dd74/platformio/package/manager/platform.py#L72
https://github.com/platformio/platformio-core/blob/65e67b64bd8d2e06e7fd9b922bb815789f46dd74/platformio/package/manager/platform.py#L86

Still, would it make sense to erase only the older packages instead of everything not contained in the platform manifest? atm it purges anything installed manually through `platform_packages = ...` overrides. (Or, at least take project config into an account when cwd inside of the project directory?)

ivankravets:
Thanks for the PR. I resolved this issue with https://github.com/platformio/platformio-core/commit/ef6e70a38b03fab62cb9f299bf0e747783387174

A little bit more clear code. In any case, we should have a solution for #3770

mcspr:
I was tracking the flow via pdb:
```diff
diff --git a/platformio/package/manager/platform.py b/platformio/package/manager/platform.py
index 7626d4d9..0a5ad4a7 100644
--- a/platformio/package/manager/platform.py
+++ b/platformio/package/manager/platform.py
@@ -148,6 +148,8 @@ class PlatformPackageManager(BasePackageManager):  # pylint: disable=too-many-an
             ]
             if any(skip_conds):
                 continue
+            import pdb
+            pdb.set_trace()
             try:
                 pm.uninstall(pkg.metadata.spec)
             except UnknownPackageError:
```

ef6e70a still does try to delete the toolchain package (what I tried to do via package owner check)
```
> c:\users\maxim\documents\platformio\platformio-core\platformio\package\manager\platform.py(153)autoremove_packages()
-> try:
(Pdb) pkg
PackageItem <path=C:\Users\maxim\.platformio\packages\toolchain-xtensa metadata=PackageMetaData <type=tool name=toolchain-xtensa version=5.100200.201223 spec={'owner': 'mcspr', 'id': 11401, 'name': 'toolchain-xtensa', 'requirements': None, 'url': None}
```

ivankravets:
See latest commits: https://github.com/platformio/platformio-core/commits/develop

We removed a code for automatic package removal. Now, everyone will personally decide when to run `pio system prune` command.

We will add a new setting where PIO will check for "unnecessary" packages 1 time per week and will inform users. But, it will never remove these packages automatically.

What do you think about this approach? We have a similar Linux OS for the APT package manager.

See update docs https://docs.platformio.org/en/latest/core/userguide/system/cmd_prune.html

mcspr:
I would agree, but not a big fan of the reminders though. I wonder if it should match certain thresholds for size / amount of packages / amount of other stuff, but not the periods of time?

Also note that it still wants to remove the custom package:
```
> pio system prune --dry-run
Dry run mode (do not prune, only show data that will be removed)

Prune cached data:
 - cached API requests
 - cached package downloads
 - temporary data
Space on disk: 494.63MB

Prune unnecessary core packages:
Calculating...
Space on disk: 0B

Prune unnecessary development platform packages:
Calculating...
Package                 Version          Size
----------------------  ---------------  --------
mcspr/toolchain-xtensa  5.100200.201223  215.25MB
Space on disk: 215.25MB

Total reclaimed space: 709.88MB
```

ivankravets:
>  I wonder if it should match certain thresholds for size / amount of packages / amount of other stuff, but not the periods of time?

Thanks, great idea! How do you think, which threshold is good for default? 1Gb?

> Also note that it still wants to remove the custom package:

Yes, it's correct behavior. See "note" https://docs.platformio.org/en/latest/core/userguide/system/cmd_prune.html

ivankravets:
So, the default threshold is 1Gb https://github.com/platformio/platformio-core/commit/2c389ae11e53071a31a1b3bc8d3f94e4cc7435c8

mcspr:
May be ok. It depends though on most common use case e.g. judging by the size of the xtensa toolchain + framework it is already 360MB by itself, gccarm + nrf52 is around 460MB, so 1GB would trigger around the time there are around 3 outdated packages (not counting the cached files though).

> Yes, it's correct behavior. See "note" https://docs.platformio.org/en/latest/core/userguide/system/cmd_prune.html

I mean, I did indent to install it through the package overrides via the registry download. Should pm auto-create those? I still kind of like the owner check, meaning PIO itself only manages the "platformio" packages, leaving user packages alone. Both registry-downloaded or the things placed into the ~/.platformio/packages manually

ivankravets:
We don't manage only `platformio` packages. It checks packages according to the platform dependencies. 

So, if you have a dev-platform that referees to a custom package, PlatformIO will not list it for "prune".

mcspr:
I mean the packages that came from platformio platform's manifest file, I though those exclusively have owner="platformio"?

From my POV, I'd need to maintain platform fork (and supply updates) + .ini change instead of just the .ini change.

ivankravets:
Please share `platform.json` from your custom dev-platform

mcspr:
The espurna's .ini just refers to the `platform = espressif8266`, I don't have the custom platform where the package is used

ivankravets:
It seems that we talk about the same but have different in a mind. What is your proposal?

`pio system prune` keeps the system clean according to the declaration from dev-platforms and PlatformIO Coe. Custom packages, or manually copied to PlatformIO's `packages` folder are intended for removal. There is a `note` section that explains how to "hide" custom packages/folders from PlatformIO. You can override https://docs.platformio.org/en/latest/projectconf/section_platformio.html#packages-dir and PlatformIO will not touch your custom packages.

Again, the goal of `pio system prune` is to keep a PlatformIO system clean.

mcspr:
I think the gist of my issue so far is that PIO has no way of tracking that package origins when it is installed through the platform_packages = ... directive. Which is understandable, since without a platform there's no way to refer back to it (...unless package manager metadata refers to the .ini file?)

mcspr:
Not to put this into another loop, but...
Should packages installed through platform_packages = ... be tracked as real packages? i.e., add an attribute to the metadata / auto-create .piokeep /  something else? i.e. following the APT metaphor:
> Packages which you have installed explicitly via install are also never proposed for automatic removal.

Which *is* explicitly requested through the .ini declaration

Also sorry for misunderstanding the platform <-> packages relation which created the misunderstanding to begin with... and I'd open another issue, but most of the discussion is already here.

ivankravets:
We are going to release PlatformIO Core 5.1 in the next few hours. Please file a new issue at https://github.com/platformio/platformio-core/issues

PR #3851, opened 2021-02-16
Bugfix for unit tests not working on x86 platform

chris-iot:
x86 used the wrong TestProcessor class and therefore complained about e.g. upload fails (and upload does not make sense when unit testing on x86). This commit fixes the most prominent bug. There still might be other parts to fix as you need to manually set "test_transport = native" for it to work, but I didn't figure out where to fix this...


ivankravets:
Thanks for the PR!

chris-iot:
Tested after upgrading to dev, works like a charm. Thanks a lot!

PR #3918, opened 2021-04-02
Explicitly specify PROGSUFFIX when compiling final binary

### LINKED ISSUE #3906: Using dot `.` in PROGNAME causes unexpected behavior during building
I want to rename the output `firmware.bin` file according to the example supplied by [the official documentation](https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html#custom-firmware-program-name).

#### platformio.ini:
```
[env:env_custom_prog_name]
platform = espressif8266
board = nodemcuv2
framework = arduino
build_flags = -D VERSION=13
extra_scripts = pre:extra_script.py
```

#### extra_script.py:

```
Import("env")

my_flags = env.ParseFlags(env['BUILD_FLAGS'])
defines = {k: v for (k, v) in my_flags.get("CPPDEFINES")}
# print(defines)

env.Replace(PROGNAME="firmware_%s" % defines.get("VERSION"))
```

The default empty `main.cpp` file is sufficient to reproduce the problem.

The previous default example will work. **However, as soon as I use the `PROGNAME` `1.2.3+4` (which includes dots), the program does not build properly anymore. It removes the extension of the `,elf` file, and the `partitions.bin` is missing (which causes fail to upload).**

So in other words, for the example above, the problem is reproducible as soon as I use the following semantic `VERSION` instead of `13`: **`1.2.3+4`**

The problem happens exactly because there is a dot inside the the `PROGNAME`. Could you please check it?

valeros:
Resolves #3906


MohammedNoureldin:
Thanks a lot @valeros! As soon as this is merged, can I somehow get a nighty version that includes this fix?

PR #3991, opened 2021-07-01
Bump zeroconf to 0.32

OttoWinter:
The [`zeroconf`](https://github.com/jstasiak/python-zeroconf) package has been upgraded to 0.32.0 recently with some bugfixes, nice work on better compliance with RFCs, and asyncio support.

Changelog: https://github.com/jstasiak/python-zeroconf#0320

Tested using: `pio device list --mdns` and quite a few mDNS-enabled devices on the network; no difference before/after upgrade

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ minimal_requirements = [
 ]
 
 if not PY2:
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR!

PR #4001, opened 2021-07-07
Feature/azure devops default prefix

Granjow:
When cloning from Azure DevOps with the HTTPS URL copied from the *clone* tab, installation fails because PlatformIO downloads the HTML file instead of using the HTTPS protocol.

This PR prefixes Azure DevOps URLs with `+git`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,3 +19,16 @@ To get started, <a href="https://cla-assistant.io/platformio/platformio-core">si
 9. Build documentation `tox -e docs` (creates a directory _build under docs where you can find the html)
 10. Commit changes to your forked repository
 11. Submit a Pull Request on GitHub.
+
+## Debugging in the IDE
+
+### PyCharm
+
+After running `tox` as above:
+
+1. In the Python Interpreter settings, add the Python interpreter that was installed with `tox`
+   with
+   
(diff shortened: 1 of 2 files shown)


ivankravets:
Could you check https://docs.platformio.org/en/latest/core/userguide/platforms/cmd_install.html#git ?

Granjow:
@ivankravets Missed this content so far – thanks for the link.

In that case I suggest as alternative to output a hint that `git+` (or similar) might be missing when unpacking from an HTTPS? url fails. Would you agree with that change?

ivankravets:
How does look the final Azure DevOps URL? Do they provide a normalized URL? We try to be service-independent. That is a reason why we created documentation and described how to use it with other source code platforms. See https://github.com/platformio/platformio-core/blob/develop/tests/package/test_meta.py#L134

Granjow:
It is of the form `https://org@dev.azure.com/org/project/_git/repo`.

Yes, I see that and it makes sense to me, therefore outputting a warning would probably help more users to fix the URL than just Azure DevOps users.

ivankravets:
This is the first case for the last 7 years. Let us see in the future. If someone will report similar issue, we implement a special autodetection for the Azure DevOps service. 

P.S: Thanks for using PlatformIO and your PR! 🙏

PR #4058, opened 2021-09-20
serialplotter filter

yhur:
I'm publishing a package called serialPlotter which the serial data just like the Arduino IDE's serial plotter.

And this pull request is an addition of new pio device monitor filter that invokes the serialPlotter so that the PIO developers  can have the Arduino IDE's serial plotter equivalent.

Regards,

### DIFF
--- a/platformio/commands/device/filters/plotter.py
+++ b/platformio/commands/device/filters/plotter.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2014-present PlatformIO <contact@platformio.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# This requires the ~/platformio/package/tool-serialplotter package
+#
+from platformio.commands.device import DeviceMonitorFilter
+import subprocess, socket, os
+
+PORT = 19200
+
+class SerialPlotter(DeviceMonitorFilter):
+    NAME = "serial_plotter"
+
+    def __init__(self, *args, **kwargs):
+        super(SerialPlotter, self).__init__(*args, **kwargs)
+        self.buffer = ''
+        self.plotter_py = os.path.expanduser('~') + '/.platformio/packages/tool-serialplotter/serialPlotter.py'
+        self.plot = subprocess.Popen(['python', self.plotter_py, '-s', str(PORT)])
+        self.plot_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+    def __call__(self):
+        try:
+            self.plot_sock.connect(('localhost', PORT))
+        except socket.error:
+            pass
+        return self
+
+    def __del__(self):
+        self.plot.kill()
+    
+    def rx(self, text):
+        self.buffer += text
+        if '\n' in self.buffer:
+            try:
+                self.plot_sock.send(bytes(self.buffer, 'utf-8'))
+            except BrokenPipeError:
+                try:
+                    self.plot_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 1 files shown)


yhur:
Needs some fix, so I'm closing the PR.

yhur:
> I'm publishing a package called serialPlotter which the serial data just like the Arduino IDE's serial plotter.
> 
> And this pull request is an addition of new pio device monitor filter that invokes the serialPlotter so that the PIO developers can have the Arduino IDE's serial plotter equivalent.
> 
> Regards,

I'm reopening the PR after fixing the cleanup procedure.

The usage of this function is as follows.
<pre>
pio device monitor -f serial_plotter 
</pre>

And I would like to get some advice on how to get the prereq(https://github.com/yhur/PIOSerialPlotter.git) installed automatically when the user invokes above command if possible.

Thanks.

yhur:
> And I would like to get some advice on how to get the prereq(https://github.com/yhur/PIOSerialPlotter.git) installed 
> automatically when the user invokes above command for the first time if possible.

For now, I put the preventive code to handle the case the serialPotter is not installed yet by the commit 'handling for the case no serialPlotter installed.

platformio/commands/device/filters/plotter.py
```diff
@@ -0,0 +1,66 @@
...
+
+PORT = 19200
+
+class SerialPlotter(DeviceMonitorFilter):
+    NAME = "serial_plotter"
+
+    def __init__(self, *args, **kwargs):
+        super(SerialPlotter, self).__init__(*args, **kwargs)
+        self.buffer = ''
+        self.plotter_py = os.path.expanduser('~') + '/.platformio/packages/tool-serialplotter/serialPlotter.py'
```
maxgerhardt, line 31: Will this work for Windows users? Better not use `os.path.join()` to concatenate paths?

yhur: Hi @maxgerhardt ,
Thank you for the suggestion. I'll modify the code on my branch for this.

maxgerhardt:
This needs the help of @ivankravets to integrate and auto-install your suggested `tool-serialplotter` package by default in PlatformIO.

yhur:
> This needs the help of @ivankravets to integrate and auto-install your suggested `tool-serialplotter` package by default in PlatformIO.

yes @maxgerhardt , I had a talk with @ivankravets already, and am working on my tool-serialplotter as a pip-installable. I will come back to you and him after working on that too. Thanks.

platformio/commands/device/filters/plotter.py
maxgerhardt: Oh, in fact it can't be assumed that PlatformIO is even installed in that path since it can be arbitrary. You should use `get_project_core_dir()` per https://github.com/platformio/platformio-core/blob/311e10f91e825348df97c0701f575f2b42260b57/platformio/debug/config/base.py#L211

yhur: Great!  Actually that's what I really needed. Thanks.

yhur: By the way, since I started changing the serialPlotter tool itself after getting @ivankravets's advice, I may change some part of this filter to take the tools modification. And I may work on it this Weekend. So let's continue this thread early next week. Thanks.

yhur: Hi @ivankravets and @maxgerhardt ,

I reworked the serialplotter python command and it’s renamed to arduplot taking @ivankravets idea of naming.
Github repo → https://github.com/yhur/arduplot

1. And I pushed it to pypi.org, so it can be installed with pip.
2. this serial_plotter filter is also updated.

Would you take a look and give me your feedback before the merge?

yhur:
Hi @ivankravets and @maxgerhardt,
I realized I overlooked the difference of the loaders of Windows and other Unix like OSes which uses the first couple of bytes of the files to determine the loading and invoking of the program while Windows depends on the extension.

When researched other packages like ‘twine’, it installs twine.exe for Windows while twine python script for MacOS/Linux. So I think I need to make ‘arduplot’ as ‘arduplot.exe’ for Windows, but don’t know how to do it. 

Could you help on this or give me some pointers to refer and to learn? Thanks.

yhur:
Hi @ivankravets and @maxgerhardt,

As an alternative resolution to the 'arduplot' command for Windows, I wrapped it with 'arduplot.cmd' batch script.
And since the terminating of the monitor and/or plot window is different between Windows and MacOS/Linux, some change was also added. I'm going to pause here and wait for your feedback. Thanks.

PR #4093, opened 2021-10-30
Add "inc" as sign that it's the root of the library

maxgerhardt:
Extends the list of folder names which PlatformIO detects as root folder in case there is no library manifests by the element "inc". 

Would fix the issue described in https://community.platformio.org/t/wrong-install-by-git-link-of-lib-without-manifest/24228.


ivankravets:
Thanks for the PR!

Issue #4099, opened 2021-11-04
Failure to build/grab dependencies when building from network share


bizzehdee:
I am trying to have my projects hosted on network shared SSD using VSCode 1.61.2 and PlatformIO (Core 5.2.2 and Home 3.3.4) on Windows 10.

I have already tried the DisableUNCCheck reg patch and it has made no difference.

### Error Log

> Executing task: C:\Users\darren\.platformio\penv\Scripts\platformio.exe run --environment LPC1768 <

Processing LPC1768 (platform: https://github.com/p3p/pio-nxplpc-arduino-lpc176x/archive/0.1.3.zip; board: nxp_lpc1768; framework: arduino)
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------Error: Traceback (most recent call last):
  File "C:\Program Files\Python39\lib\ntpath.py", line 647, in realpath
    path = _getfinalpathname(path)
FileNotFoundError: [WinError 161] The specified path is invalid: '\\\\VAULT\\Work_Darren\\Marlin\\https:\\github.com\\p3p\\pio-nxplpc-arduino-lpc176x\\archive\\0.1.3.zip'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\darren\.platformio\penv\lib\site-packages\platformio\__main__.py", line 129, in main
    cli()  # pylint: disable=no-value-for-parameter
  File "C:\Users\darren\.platformio\penv\lib\site-packages\click\core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\darren\.platformio\penv\lib\site-packages\click\core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "C:\Users\darren\.platformio\penv\lib\site-packages\platformio\commands\__init__.py", line 44, in invoke
    return super(PlatformioCLI, self).invoke(ctx)
  File "C:\Users\darren\.platformio\penv\lib\site-packages\click\core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\darren\.platformio\penv\lib\site-packages\click\core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\darren\.platformio\penv\lib\site-packages\click\core.py", line 754, in invoke
    return __callback(*args, **kwargs)
[...]

PR #4106, opened 2021-11-10
Fix network storage build support on windows

bizzehdee:
realpath and abspath are only interchangable on unix OS's. 

abspath returns a normalized absolutized version of the pathname, whereas realpath just parses through symbolic links to paths.

This breaks when \\networkpath\work is mapped to X: or any other drive letter on windows but as all paths work the same under all unix based OS's, it is unaffected.


bizzehdee:
Fixes issue #4099

ivankravets:
Thanks for the PR! Resolved in https://github.com/platformio/platformio-core/commit/4687665ff3706f02fae8cd1ea220ae502a4f6ec9

PR #4153, opened 2022-01-09
Unix line-endings for extensions.json

thinkyhead:
I'm now including `extensions.json` in my repository so that users will get a message about the recommended extensions.

Every so often the file is written with CRLF line endings, and then at other times it is written with LF line endings. I'd prefer that it is always written with the same (LF) line endings so that it doesn't appear to be "changed" according to git.

Simply applying Unix line-endings to the template seems to prevent the issue.


ivankravets:
Thanks for the PR! 🙏

PR #4210, opened 2022-03-30
Don't append --checks=* when the --config or --config-file flags are set

### LINKED ISSUE #4186: Enable "--config-file" flag for clang-tidy
What kind of issue is this?

- [ x] **Feature Request**.

*You can erase any parts of this template not applicable to your Issue.*

### Configuration

**Operating system**: Win10

**PlatformIO Version** (`platformio --version`): latest

### Description of problem
As a developer I would like to tell clang-tidy to use an external config file to keep my configs clean. Especially when handling multiple clang-tidy checks the code looks messy if I am forced to use the clangtidy extra flag --checks=<string> for that purpose.
Instead it would be great to use a config file that is made of yaml like this:

    ---
    Checks:              '-*,some-check'
    WarningsAsErrors:    ''
    HeaderFilterRegex:   ''
    FormatStyle:         none
    InheritParentConfig: true
    User:                user
    CheckOptions:
      - key:             some-check.SomeOption
        value:           'some value'
    ...

CLang Tidy comes with a flag --config-file for that scenario (see clang-tidy --help, it is not documented on the websites). Its help text is: "Specify the path of .clang-tidy or custom config file: e.g. --config-file=/some/path/myTidyConfigFile This option internally works exactly the same way as --config option after reading specified config file. Use either --config-file or --config, not both."

Actually it is not possible to use this flag because at least the Checks part is overriden due to the logic in [https://github.com/platformio/platformio-core/blob/4839fe37a3d1433737c80147ce9a205100dc4925/platformio/commands/check/tools/clangtidy.py](url):

```
if not self.is_flag_set("--checks", flags):
            cmd.append("--checks=*")
```

monsterzack5:
Fixes #4186

Appending --checks=* causes clang-tidy to ignore the flags --config and --config-file from `check_flags`, which breaks the ability to use a clang-tidy file.

### DIFF
--- a/platformio/commands/check/tools/clangtidy.py
+++ b/platformio/commands/check/tools/clangtidy.py
@@ -60,7 +60,9 @@ class ClangtidyCheckTool(CheckToolBase):
 
         cmd = [tool_path, "--quiet"]
         flags = self.get_flags("clangtidy")
-        if not self.is_flag_set("--checks", flags):
+        if not (
+            self.is_flag_set("--checks", flags) or self.is_flag_set("--config", flags)
+        ):
             cmd.append("--checks=*")
 
         project_files = self.get_project_target_files(self.options["patterns"])


ivankravets:
Thanks for the PR! 🙏

PR #4217, opened 2022-04-08
Avoid working with detached / non-existent git branches when checking for updates

### LINKED ISSUE #4139: Actual git version is not displayed for `platform_packages`
What kind of issue is this?

- [x] **PlatformIO Core**.

### Configuration

**Operating system**:
Fedora Linux 35 (Container Image)

**PlatformIO Version** (`platformio --version`):
PlatformIO Core, version 5.2.5a1

### Description of problem

Given the following platformio.ini
```ini
[env]
platform = espressif8266@3.2.0
framework = arduino
platform_packages =
    platformio/framework-arduinoespressif8266 @ git+https://github.com/esp8266/Arduino.git
; notice that there's no version pin, it fetches the master branch

[env:d1_mini]
board = d1_mini
```

Extra package override is successfully fetched into `~/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/` and reported as version `3.1.0-dev+sha.42aa0e6d`
However, manually changing the commit by using `git checkout 55ef3e73978ad010d44ad7e4c9bd4b4f32455989` does not seem to affect the version displayed in the `PACKAGES` list when doing `pio run`, and instead the `.git/.piopm` metadata file "version" field is used (from the installation time).

```
$ pio run
Processing d1_mini (platform: espressif8266@3.2.0; board: d1_mini; framework: arduino)
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/espressif8266/esp12e.html
PLATFORM: Espressif 8266 (3.2.0) > Espressif ESP8266 ESP-12E
HARDWARE: ESP8266 80MHz, 80KB RAM, 4MB Flash
PACKAGES:
 - framework-arduinoespressif8266 3.1.0-dev+sha.42aa0e6d
 - tool-esptool 1.413.0 (4.13)
 - tool-esptoolpy 1.30000.201119 (3.0.0)
 - toolchain-xtensa 2.100300.210717 (10.3.0)
^CBuild interrupted.
$ git -C ~/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910/ rev-parse HEAD
55ef3e73978ad010d44ad7e4c9bd4b4f32455989
```

Shouldn't the package manager run `git rev-parse HEAD` (or track `.git/HEAD` changes) and use that value instead of version saved in the `.piopm`? Or, at least attempt to refresh it. I have noticed inconsistent version reports at the esp8266/Arduino repo when user has master-branch package configured, it would be nice to know the actual revision of the repo without asking the user to go to the ~/.platformio/packages dir and run `git` command manually :)

(and I assume the same behaviour for `lib_deps`, but I have not tried it)

mcspr:
b/c we can't use `pull` anyway in that situation
also ask *only* for the specific branch via `refs/heads/{branch}` and fail when it is not available. `pkg update` / `pkg outdated` will not do or display anything if that's the case.

resolves #4139 
(and I am not familiar with the svn & hg cases, so only git implementation was changed)

### DIFF
--- a/platformio/package/manager/_update.py
+++ b/platformio/package/manager/_update.py
@@ -72,9 +72,14 @@ class PackageManagerUpdateMixin(object):
             return None
         if not vcs.can_be_updated:
             return None
+
+        vcs_revision = vcs.get_latest_revision()
+        if not vcs_revision:
+            return None
+
         return str(
             self.build_metadata(
-                pkg.path, pkg.metadata.spec, vcs_revision=vcs.get_latest_revision()
+                pkg.path, pkg.metadata.spec, vcs_revision=vcs_revision
             ).version
         )
 
--- a/platformio/package/vcsclient.py
+++ b/platformio/package/vcsclient.py
@@ -217,12 +217,17 @@ class GitClient(VCSClientBase):
             return self.get_current_revision()
         branch = self.get_current_branch()
         if not branch:
-            return self.get_current_revision()
-        result = self.get_cmd_output(["ls-remote"])
-        for line in result.split("\n"):
-            ref_pos = line.strip().find("refs/heads/" + branch)
-            if ref_pos > 0:
-                return line[:ref_pos].strip()[:7]
+            return None
+
+        branch_ref = f"refs/heads/{branch}"
+        result = self.get_cmd_output(["ls-remote", self.remote_url, branch_ref])
+
+        if result:
+            for line in result.split("\n"):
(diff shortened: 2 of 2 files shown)


ivankravets:
Thanks for the PR!

PR #4262, opened 2022-05-14
qtcreator IDE template now generates a "generic" Qt project

### LINKED ISSUE #3046: Qt Creator integration enhancement request

belm0:
Change the `qtcreator` IDE template to generate a [Qt Creator generic project](https://doc.qt.io/qtcreator/creator-project-generic.html)

(In contrast, the existing `qtcreator` template creates a native project having dependencies on Qt runtime and the `qmake` tool.)

Fixes #3046

Derived from original PR #3072


ivankravets:
Thanks for the PR! So, the main question is still actual. Why we can't merge this PR to https://github.com/platformio/platformio-core/tree/develop/platformio/project/tpls/qtcreator ? Why do we need 2 different "IDEs"? Does it make sense to remove [platformio.pro.tpl](https://github.com/platformio/platformio-core/blob/develop/platformio/project/tpls/qtcreator/platformio.pro.tpl) and replace with this PR?

belm0:
> So, the main question is still actual. Why we can't merge this PR to https://github.com/platformio/platformio-core/tree/develop/platformio/project/tpls/qtcreator ? Why do we need 2 different "IDEs"? Does it make sense to remove [platformio.pro.tpl](https://github.com/platformio/platformio-core/blob/develop/platformio/project/tpls/qtcreator/platformio.pro.tpl) and replace with this PR?

I'm new to Qt Creator, but it seems to have two very different ways to configure a project.  Originally this IDE was dedicated to Qt application development, and used `.pro` file.  Later, a more general configuration mechanism was added for non-Qt projects.

I don't know if some platform IO projects might use Qt-- I guess not?

So it's up to your policy about migration.  Yes, we could replace `qtcreator` with this new implementation.  But any published tutorials or documentation on plotformio Qt Creator integration would instantly become non-working.

ivankravets:
> I don't know if some platform IO projects might use Qt-- I guess not?

No, PlatformIO does not compatible with QT's ecosystem and does not depend on it.

> Later, a more general configuration mechanism was added for non-Qt projects.

This is exactly what we need.

>  But any published tutorials or documentation on plotformio Qt Creator integration would instantly become non-working.

As we understand, a few people might use Qt in pair with PlatformIO. We even don't know what is the reason for this. It will be still limited integration solution in comparison to VSCode.

P.S: Does this PR work for you? What was your motivation to use Qt instead of VSCode?

belm0:
I tested this PR with a hello world on ESP32.

Output file contents look this:

<details><summary>Makefile</summary>

```
all:
        platformio -c qtcreator run

# forward any other target (clean, build, etc.) to pio run
%:
        platformio -c qtcreator run --target $*
```
</details>

<details><summary>platformio.cflags</summary>

```
-mlong-calls -Wno-frame-address -std=gnu99 -Wno-old-style-declaration -ffunction-sections -fdata-sections -Wno-error=unused-function -Wno-error=unused-variable -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-sign-compare -ggdb -Os -freorder-blocks -Wwrite-strings -fstack-protector -fstrict-volatile-bitfields -Wno-error=unused-but-set-variable -fno-jump-tables -fno-tree-switch-conversion -MMD
```
</details>

<details><summary>platformio.config</summary>

```
#define PLATFORMIO 60000
#define ARDUINO_M5STACK_Core2
#define BOARD_HAS_PSRAM
#define HAVE_CONFIG_H
#define MBEDTLS_CONFIG_FILE "mbedtls/esp_config.h"
#define UNITY_INCLUDE_CONFIG_H
#define WITH_POSIX
#define _GNU_SOURCE
#define IDF_VER "v4.4-beta1-189-ga79dc75f0a"
#define ESP_PLATFORM
#define _POSIX_READER_WRITER_LOCKS
#define ARDUINO_ARCH_ESP32
#define ESP32
#define F_CPU 240000000L
#define ARDUINO 10812
#define ARDUINO_VARIANT "m5stack_core2"
#define ARDUINO_BOARD "M5Stack Core2"
#define ARDUINO_PARTITION_default
```
</details>

<details><summary>platformio.creator</summary>

```
[General]
```
</details>

<details><summary>platformio.cxxflags</summary>

```
-mlong-calls -Wno-frame-address -std=gnu++11 -fexceptions -fno-rtti -ffunction-sections -fdata-sections -Wno-error=unused-function -Wno-error=unused-variable -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-sign-compare -ggdb -Os -freorder-blocks -Wwrite-strings -fstack-protector -fstrict-volatile-bitfields -Wno-error=unused-but-set-variable -fno-jump-tables -fno-tree-switch-conversion -MMD
```
</details>

<details><summary>platformio.files</summary>

```
Makefile
platformio.ini
src/main.cpp
```
</details>

<details><summary>platformio.includes</summary>

```
./
/Users/john/Documents/Arduino/platformio_test2/include
/Users/john/Documents/Arduino/platformio_test2/src
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/config
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/newlib/platform_include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/freertos/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/freertos/include/esp_additions/freertos
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/freertos/port/xtensa/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/freertos/include/esp_additions
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_hw_support/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_hw_support/include/soc
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_hw_support/include/soc/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_hw_support/port/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/heap/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/log/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/lwip/include/apps
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/lwip/include/apps/sntp
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/lwip/lwip/src/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/lwip/port/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/lwip/port/esp32/include/arch
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/soc/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/soc/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/soc/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/hal/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/hal/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/hal/platform_port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_rom/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_rom/include/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_rom/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_common/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_system/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_system/port/soc
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_system/port/public_compat
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/xtensa/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/xtensa/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/driver/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/driver/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_pm/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_ringbuf/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/efuse/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/efuse/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/vfs/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_wifi/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_event/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_netif/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_eth/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/tcpip_adapter/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_phy/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_phy/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_ipc/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/app_trace/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_timer/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/mbedtls/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/mbedtls/mbedtls/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/mbedtls/esp_crt_bundle/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/app_update/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/spi_flash/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bootloader_support/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/nvs_flash/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/pthread/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_gdbstub/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_gdbstub/xtensa
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_gdbstub/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/espcoredump/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/espcoredump/include/port/xtensa
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/wpa_supplicant/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/wpa_supplicant/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/wpa_supplicant/esp_supplicant/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/ieee802154/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/console
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/asio/asio/asio/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/asio/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bt/common/osi/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bt/include/esp32/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bt/common/api/include/api
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bt/common/btc/profile/esp/blufi/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bt/common/btc/profile/esp/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/bt/host/bluedroid/api/include/api
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/cbor/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/unity/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/unity/unity/src
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/cmock/CMock/src
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/coap/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/coap/libcoap/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/nghttp/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/nghttp/nghttp2/lib/includes
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-tls
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-tls/esp-tls-crypto
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_adc_cal/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_hid/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/tcp_transport/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_http_client/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_http_server/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_https_ota/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_lcd/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_lcd/interface
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/protobuf-c/protobuf-c
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/protocomm/include/common
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/protocomm/include/security
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/protocomm/include/transports
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/mdns/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_local_ctrl/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/sdmmc/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_serial_slave_link/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_websocket_client/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/expat/expat/expat/lib
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/expat/port/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/wear_levelling/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/fatfs/diskio
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/fatfs/vfs
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/fatfs/src
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/freemodbus/common/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/idf_test/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/idf_test/include/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/jsmn/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/json/cJSON
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/libsodium/libsodium/src/libsodium/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/libsodium/port_include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/mqtt/esp-mqtt/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/openssl/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/perfmon/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/spiffs/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/ulp/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/wifi_provisioning/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/button/button/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/json_parser
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/json_parser/jsmn/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/json_generator
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_schedule/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_rainmaker/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/qrcode/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/ws2812_led
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_littlefs/src
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp_littlefs/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/dotprod/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/support/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/hann/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/blackman/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/blackman_harris/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/blackman_nuttall/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/nuttall/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/windows/flat_top/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/iir/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/fir/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/add/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/sub/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/mul/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/addc/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/mulc/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/math/sqrt/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/matrix/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/fft/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/dct/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/conv/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/common/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/kalman/ekf/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-dsp/modules/kalman/ekf_imu13states/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/tool
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/typedef
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/image
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/math
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/nn
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/layer
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/detect
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp-face/include/model_zoo
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp32-camera/driver/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/esp32-camera/conversions/include
/Users/john/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32/include/fb_gfx/include
/Users/john/.platformio/packages/framework-arduinoespressif32/cores/esp32
/Users/john/.platformio/packages/framework-arduinoespressif32/variants/m5stack_core2
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/ArduinoOTA/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/AsyncUDP/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/BLE/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/BluetoothSerial/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/DNSServer/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/EEPROM/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/ESP32/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/ESPmDNS/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/Ethernet/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/FFat/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/FS/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/HTTPClient/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/HTTPUpdate/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/HTTPUpdateServer/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/LittleFS/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/NetBIOS/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/Preferences/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/RainMaker/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/SD/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/SD_MMC/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/SPI/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/SPIFFS/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/SimpleBLE/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/Ticker/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/USB/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/Update/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/WebServer/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/WiFi/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/WiFiClientSecure/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/WiFiProv/src
/Users/john/.platformio/packages/framework-arduinoespressif32/libraries/Wire/src
/Users/john/.platformio/packages/toolchain-xtensa-esp32/xtensa-esp32-elf/include/c++/8.4.0
/Users/john/.platformio/packages/toolchain-xtensa-esp32/xtensa-esp32-elf/include/c++/8.4.0/xtensa-esp32-elf
/Users/john/.platformio/packages/toolchain-xtensa-esp32/lib/gcc/xtensa-esp32-elf/8.4.0/include
/Users/john/.platformio/packages/toolchain-xtensa-esp32/lib/gcc/xtensa-esp32-elf/8.4.0/include-fixed
/Users/john/.platformio/packages/toolchain-xtensa-esp32/xtensa-esp32-elf/include
```
</details>

<details><summary>platformio.ini</summary>

```
; PlatformIO Project Configuration File
;
;   Build options: build flags, source filter
;   Upload options: custom upload port, speed and extra flags
;   Library options: dependencies, extra library storages
;   Advanced options: extra scripting
;
; Please visit documentation for the other options and examples
; https://docs.platformio.org/page/projectconf.html

[env:m5stack-core2]
platform = espressif32
board = m5stack-core2
framework = arduino
```
</details>

ivankravets:
What was your motivation to use Qt instead of VSCode?

belm0:
> As we understand, a few people might use Qt in pair with PlatformIO. We even don't know what is the reason for this. It will be still limited integration solution in comparison to VSCode.
> 
> P.S: Does this PR work for you? What was your motivation to use Qt instead of VSCode?

Yes, it's working as far as I can tell (no important warnings from IDE in the code, IDE autocomplete works, build works).

I guess I have a bias so I don't consider using VSCode.  My first choice was CLion since I'm very familiar with PyCharm already, but I develop on an old laptop and I heard CLion can be slow (not to mention expensive).  (Maybe I'd have a speed problem with VSCode too.)  I'm not really 

Qt Creator seems incredibly efficient-- my laptop fan doesn't even start when using it.

I agree about just replacing `qtcreator`, but updating the platformio-docs would become a priority.

ivankravets:
If you could provide a PR to https://github.com/platformio/platformio-docs/blob/develop/integration/ide/qtcreator.rst , we would be happy to merge both PRs.

See https://github.com/platformio/platformio-core/blob/develop/CONTRIBUTING.md

belm0:
perhaps the tests don't work on macOS

Would you please approve running the workflow on this PR so that it can be tested by CI?

```
[gw1] [ 11%] FAILED tests/commands/test_ci.py::test_ci_boards

==================================== FAILURES =====================================
_________________________________ test_ci_boards __________________________________
[gw1] darwin -- Python 3.7.7 /Users/john/tmp/platformio-core/.tox/py37/bin/python

clirunner = <click.testing.CliRunner object at 0x10deb7fd0>
validate_cliresult = <function validate_cliresult.<locals>.decorator at 0x10e0797a0>

    def test_ci_boards(clirunner, validate_cliresult):
        result = clirunner.invoke(
            cmd_ci,
            [
                join("examples", "wiring-blink", "src", "main.cpp"),
                "-b",
                "uno",
                "-b",
                "leonardo",
            ],
        )
>       validate_cliresult(result)

tests/commands/test_ci.py:38:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

result = <Result SystemExit(2)>

    def decorator(result):
>       assert result.exit_code == 0, "{} => {}".format(result.exception, result.output)
E       AssertionError: 2 => Usage: ci [OPTIONS] [SRC]...
E         Try 'ci --help' for help.
E
E         Error: Invalid value for '[SRC]...': Found invalid path: examples/wiring-blink/src/main.cpp
E
E       assert 2 == 0
E        +  where 2 = <Result SystemExit(2)>.exit_code

tests/conftest.py:33: AssertionError
```

ivankravets:
Did you use `git clone --recursive`?

belm0:
Thank you.

I think the ref for the doc submodule needs to be updated.  Shall I open another PR?

ivankravets:
No, no. I'll update together with the next commit.

dewhisna:
Thanks @belm0 for getting my original PR revived and things completed!

And yes, your description is correct.  Qt Creator has two modes (actually three now).  It has .pro files (which use its QMake build system) and generic project files (with a raw Makefile). And most recently, they've added full support for CMake.  In fact, CMake will soon completely displace QMake and .pro files entirely.

The .pro files are specific to the, now deprecated QMake, and this original PR was intended to switch PlatformIO projects to use the generic Makefile mechanism of Qt Creator.  It was kept independent of the original qtcreator target, which targeted .pro files, in case users still wanted to use QMake and kept existing documentation for .pro files correct.  Though we should probably start looking at CMake now, since that's where Qt is heading.

And @ivankravets , as for "why Qt Creator and not VSCode"?  The simple answer, at least for me, is that I absolutely can't stand VSCode; it drives me bonkers.  Most days, I would rather use plain-old 'vi' from a terminal shell prompt than VSCode.  Plus, I do a lot of Qt development work for the company I work for and having a single IDE for everything makes my life easier.  It's one of the reasons, I assume, that PlatformIO had the qtcreator IDE integration scripts already.

While PlatformIO is nicely integrated with VSCode, it shouldn't be completely married to it and should remain as IDE-agnostic as it can.  I see PlatformIO as the source-control fix that the entire Arduino and embedded world has been desperately needing for a long time.  It's like the "Conan" for it (and interestingly, in Python as well).  And most of the time, for building projects, I'll just use the CLI form of PlatformIO.

Before PlatformIO, the whole Arduino world was a completely chaotic mess of people randomly compiling code from whatever they had lying around at the time from random (unknown) sources with no way to replicate anything and heaven forbid them being able to pick up a project five years later on a new computer and compile and get the same original binary they had the last time it was built.  And recreating their build environment was a very manual and involved process and worked only if they had kept really good notes for every piece of the puzzle.

Being able to pull specific upstream repos and designate specific commit hashes for source and properly manage embedded projects has been a real lifesaver.  Before PlatformIO, I had started trying to cobble together a suite of CMake files that did much of the same thing.  But it was a beast to maintain and still had lots of issues.  Then, I discovered PlatformIO and now use it almost exclusively for embedded projects.

But VSCode ... well, one day it got me so frustrated and bent out of shape that I threw the computer across the room.  Well, OK, it was a bad day already and there were other contributing factors, but VSCode is what finally pushed me over the edge.

ivankravets:
Thanks for the extensive feedback! 🙏 

@dewhisna , @belm0 , could you try generating a project for CLion (  https://github.com/platformio/platformio-core/tree/develop/platformio/project/tpls/clion ) and import it in Qt? CLion uses Cmake.

dewhisna:
@ivankravets -- I just took one of my existing PlatformIO projects that targets an ATmega328PB micro, which I was previously using the generic Makefile method from this PR's history, but in a clean source folder without the other project import (so that it wouldn't interfere and affect the test), and tried it.  I was able to have PlatformIO create a `CMakeLists.txt` file successfully and it loaded and configured correctly on my Linux system in Qt Creator 6.0.2 with CMake 3.23.1.

The library structure in Creator looks correct and had all the source and header files correctly designated for the `libdeps` pieces.  When trying to build, it looked to be calling the correct `avr-g++` compiler from the `.platformio/packages/toolchain-atmelavr/bin/` directory, just as it should.

But the build failed not able to find the `Arduino.h` include, which, if I remember correctly, it should have been picking up from `.platformio/packages/framework-arduinoavr/cores/MiniCore/`.

I'm going to have to do some more digging on this to see what exactly is going on.  My guess is that it's either missing an include path somewhere or a path got set incorrectly for the platform core itself.

It may take a while for me to get back to this, since my plate is quite full at the moment with other projects.  But at a first quick glance at this, it looks very promising, now that Qt Creator has good CMake support.  In fact, the results I dare say looks better than the generic Makefile method that started this PR, as the library folder hierarchy is more complete and better organized.

I'm thinking that with a couple of minor tweaks, the `CLion` target will be a viable solution for Qt Creator as well.  And who knows, this may just be an existing bug in that template for `CLion` as well that needs fixing that I stumbled upon (or a problem with my `platformio.ini` file or my custom boards JSON or something) and it may already work out-of-the-box for Qt Creator as well as it does on CLion.

belm0:
Am I using cmake incorrectly?  When I try it, all the compile commands are missing the includes entirely.

```
$ pio project init --ide clion --board m5stack-core2
$ cmake .
$ make
In file included from AXP192.cpp:1:
AXP192.h:4:10: fatal error: Wire.h: No such file or directory
 #include <Wire.h>
```

ivankravets:
You can't use CMake directly. PlatformIO generates "wrappers" around PlatformIO Core and implements them as CMake "targets". You must call them instead of CMake.

dewhisna:
Actually, it looks like the `CLion` template creates two targets `Production` and `Debug`.  So, @belm0, you can pass one of those targets to your call to `cmake` instead of letting it default to `all`.  That worked for me to get all of the paths and such.

However, instead of calling cmake manually like that, after you run the platformio init step, you can just open the `CMakeLists.txt` file directly in Qt Creator like you would any Creator project and let it do all of the calls to cmake for you.  It also can properly handle all of the shadow-build folder logic so you don't clutter up your source checkout folder either.

After you've imported and configured your project in Creator, press Ctrl-K (or click in the little command launch box in the bottom left corner) and enter `cm <target>` where `<target>` is one of the build targets created by the CLion template.  Creator will automatically show you a list of completions in the drop-list of targets from the cmake configuration.

Also, for your kits, so that Creator will configure things correctly, just create a kit for your embedded target (for example, in my test case for this, I have a "noQt-PlatformIO-AVR" kit that I setup) and select that in the configure process.  That way, Creator will know what type of compiler you're using, its ABI, details on how to launch the debugger, how to find cmake, etc.  And you can set it up without a Qt so that it knows you aren't trying to build/link any of the Qt stuff -- though Qt does support some MCUs now.

@ivankravets -- the problem I have now with the `CLion` template is this:

```cmake
add_custom_target(
    Production ALL
    COMMAND platformio -c clion run "$<$<NOT:$<CONFIG:All>>:-e${CMAKE_BUILD_TYPE}>"
    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
)

add_custom_target(
    Debug ALL
    COMMAND platformio -c clion debug "$<$<NOT:$<CONFIG:All>>:-e${CMAKE_BUILD_TYPE}>"
    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
)
```

Specifically with the `-e${CMAKE_BUILD_TYPE}>` part.  That implies that your `platformio.ini` file uses `[env:Debug]` or `[env:Release]` or `[env:RelWithDebInfo]`, etc., i.e. one of the CMake standard build types.  That was not the case for me and gave an error saying that it wasn't.

If I hand-edit the `CMakeLists.txt` and change that to be `-eATmega328PB20`, which happened to be my 20MHz ATmega328PB build environment name in my `platformio.ini`, it correctly found the core libraries, got all of the paths correct, and built just fine.

So ... my conclusion is that yes the `CLion` template works well for Qt Creator's CMake project support.  However, there's a deficiency in the `CLion` template that instead of using `${CMAKE_BUILD_TYPE}` in those custom targets, it needs to walk the list of environments in the `platformio.ini` file and create separate custom targets in the `CMakeLists.txt` corresponding to them, since users seldom use the CMake build types for their environments but instead would use specific board and system configurations for their environments.

belm0:
@dewhisna thank you, I'd like to try this when I find some time.  About the kit config, would you mind sharing some screenshots so I have an example?

dewhisna:
@belm0 -- The first step is to make sure you have 'CMake' itself setup in Creator.  Select Tools->Options->Kits->CMake tab.  And make sure Creator sees your CMake.  I'm using the system installed version in this screenshot, which is automatically detected, but you can point it to any CMake location on your system (allowing you to run different CMake versions, etc):
image

Next you can optionally set up the PlatformIO compiler.  It's optional for the actual build because the PlatformIO generated CMake should be giving CMake compiler-discovery hints for the toolchain installed on PlatformIO.  However, for things like the debugger and syntax checking and such, go to the "Compilers" tab of the Kits dialog and set up the path and details for the compiler.  Here's where I've selected the g++ compiler for the AVR.  You can set both the C and C++ compilers.  The name you use for it is somewhat arbitrary but will be used on the "Kits" tab to select the compiler for the kit:
image

Then go to the "Debuggers" tab and do the same thing for the gdb debugger.  This too is optional if you aren't going to do live debugging:
image

Then go to the "Kits" tab and create a new kit.  Select "Bare Metal Device".  Select the C and C++ compilers you set up previously or "No compiler" if you elected not to set them up.  Select the debugger or "None" if you elected not to set up a debugger.  Select "None" for the Qt version.  Make sure it shows the correct CMake.  You can also edit the CMake Configuration variables for the kit if you have any settings you want always set for this kit.  The default is things like the compiler you picked, etc.  And select the generators you want CMake to use.  I recommend and use 'Ninja' (if you've installed 'ninja-build'), but you can also use 'Unix Makefiles' on *nix systems or msbuild on Windows, etc.:
image

Then, after you have done your project init with PlatformIO and have the CMakeLists.txt file, just open it in Creator.  It will automatically guide you through selecting the kit.  Simply pick the kit you just created and click "Configure Project":
image

Normally, the default of build 'all' would work, but due to the aforementioned CLion template peculiarities, you'll need to target the specific build.  You can do a one-off build by pressing Ctrl-K and entering 'cm Production', as I mentioned earlier.  Or, you can click "Projects" on the left and set the "Build Step" to build the "Production" target, after unselecting "all".  This will become the build target it runs when you click 'build' with this configuration selected:
image

And as I was complaining about above, regarding the CMake build type and the PlatformIO env, you'll either have to name your env 'Debug' or 'Release' or other standard CMake build-type.  Or, you'll have to change the `CMAKE_BUILD_TYPE` variable to match your env.  This is done on the same Projects page.  Highlighted here is the normal default for the 'Debug' target:
image

With how the CLion template creates the CMakeLists.txt file, change this to your env name and click 'Apply Configuration Changes':
image

And that's it ...  CMake works pretty well.  As for the debugger setup, I haven't tried the debugger on the AVR targets with Creator, but I have done live debugging on an ARM STM32F746 target using the STLink interface to debug the startup code when creating my own platform core from scratch, as I had messed up the `_init()` function and was never even making it to `main()` and needed to see what was going on.  Setting up the debugger for live debugging requires a bit more setup in the Tools->Options->Debugger dialog and also very system, OS, and hardware dependent, so I won't go into those details here (use Google).  You can also use qemu to emulate the device's processor and actually run the target code on your host system -- or at least you can in Linux.  I haven't tried qemu on other operating systems.

JoseOlin:
Hi!

Totally interested in Qt Creator being better supported.

I confirm that on a laptop with Intel i3 and 8GB of RAM Qt Creator feels much faster than VS Code; I can even run it on a mini laptop with an Atom processor. That is why in medium projects I completely prefer Qt Creator and I am excited that this initiative is revived.

I only use VS Code when the priority is to collaborate with others, as it is definitely more popular and easier to configure with PlatformIO, but I think it has a lot to do with it having more support, so I am excited to see this initiative revived.

I agree that PlatformIO remains IDE agnostic, or at least that the other options appear on the landing page, because at first it seems that VS Code is the only possible option.

JoseOlin:
Regarding the new template for QtCreator, I have tested it and it works fine. The only detail is that before, in the .pro the paths to the headers of the board files were included, so you could navigate to them.

For example, having:

#define pinLED1 LED1

With Ctrl + Click on LED1 it sent me to the definition of LED1 on the board headers.  In the new template, although the project compiles and uploads correctly to the board, it does not allow me to navigate to these defines. 

Any recommendations to be able to navigate to these?

JoseOlin:
Thanks for the instructions for adding the Kit @dewhisna , I had been wondering for a long time how it would be possible to add one for PlatformIO.

How can I add the Device type: Bare Metal Device?
I managed to define the compilers and debugger but when creating the kit, in Device, Bare Metal does not appear as an option.

dewhisna:
@JoseOlin -- Oh, sorry.  To enable Bare Metal devices, you have to enable the Bare Metal support plugin in Help->About Plugins.  I didn't think about the plugin when I wrote the above details since it was already enabled on my computers and been so long since I had done a total clean install of these tools:

image


Here's the Qt documentation page:
https://doc.qt.io/qtcreator/creator-developing-baremetal.html

JoseOlin:
Thanks a lot!

belm0:
After this work I tried to take QtCreator further and use it in my workflow.  There's a decent SerialTerminal plugin.  However, I couldn't find an obvious way to stop the terminal during the deployment steps.  To me, having a smooth cycle of code change -> compile -> deploy -> monitor is essential, and I don't want worry about what the monitor state is and whether it will conflict with the upload.  Filed https://bugreports.qt.io/browse/QTCREATORBUG-27551

The vscode load on my computer is not great, but tolerable, and the compile / deploy / monitor cycle works well.

JoseOlin:
I didn't know about the SerialTerminal plugin, thanks for mentioning it, @belm0 .
Indeed, it would be ideal if the Deploy stopped it automatically. 
By the way, how do you run it on deploy?

I usually ran PlatormIO's own monitor or a handmade one based on the QSerialPort examples.

It would also be very convenient if the deploy command was automatically defined at project initialization. As far as I could see, build and clean work fine but the command for Deploy and Run still needs to be assigned by hand, right?

belm0:
> By the way, how do you run it on deploy?

I don't know of a way to run SerialTerminal automatically on deploy.

I guess it would be fine to forgo SerialTerminal and have a make target to deploy & monitor (via platformio).  I haven't learned enough about Qt Creator to map that to the UI though.

It would be a small change to the Makefile template to add a deploy_and_monitor target.

JoseOlin:
Oh sure, something like this:

```
all:
	platformio -c qtcreator run

deploy:
	platformio run --target upload

# regenerate project files to reflect platformio.ini changes
project-update:
	@echo "This will overwrite project metadata files.  Are you sure? [y/N] " \
	    && read ans && [ $${ans:-'N'} = 'y' ]
	platformio project init --ide qtcreator

# forward any other target (clean, build, etc.) to pio run
%:
	platformio -c qtcreator run --target $*
``` 

This way we only need to define the `deploy` argument, which looks like it could be handled like `all` and `clean`.


ksnip_20220522-170511

Setting the Run Executable and arguments by hand is still required; it could be added to the deploy rule in the Makefile, if we find some way to avoid this problem.

ksnip_20220522-170015
 
I stay tuned.

Happy to have found the PlatformIO+QtCreator team ;)

dewhisna:
@JoseOlin -- from what I understand about that `ioctl error`, it's because the process isn't running in a terminal environment, which makes sense if Creator didn't set up a terminal for it to run in.  I'm curious ...  what happens if you change the executable to something that provides that environment and spawns platformio from there?

I see two potential things to try -- one is to just try `bash` and see if that alone is sufficient, by setting the executable to `bash` (not sure if just `bash` will work or if you need to enter the full path to `bash`, like `/usr/bin/bash`) and passing `-c platformio device monitor` where it says 'Command line arguments'.

That may or may not work, since a bash shell isn't a full terminal, but it might decouple it enough from Creator that it's worth a try.  If it doesn't, then the second thing to try is to launch something like `xterm` (assuming you first `sudo apt-get install xterm` if you don't have it already installed), setting the executable to `xterm` (or `/usr/bin/xterm` if it wants the full path).  And setting `-hold -e platformio device monitor` for the 'Command line arguments'.  The `-hold` argument will keep the xterm window up after it finishes and exits so you can see what happened.

I'm betting that one or both of those will eliminate that ioctl error...  though I haven't actually tried it, since I don't have any devices or programming hardware here at the moment.  There may even be an option on Creator for launching xterm for you.  I know it's got that for desktop apps, particularly for running and debugging command-line applications.  I can't remember right off if it's there for all targets or just desktop targets -- if it's only the latter, you might can change your kit from 'bare metal' back to desktop and let `platformio` be the executable for your application.  In any case, look around in Creator for things about running targets using xterm, as it may simplify setting things up.

JoseOlin:
Yes!  Thanks @dewhisna !

It works with xterm

ksnip_20220523-140118


The Makefile would look something like this:

```
all:
	platformio -c qtcreator run

deploy:
	platformio run --target upload
	xterm -hold -e platformio device monitor

# regenerate project files to reflect platformio.ini changes
project-update:
	@echo "This will overwrite project metadata files.  Are you sure? [y/N] " \
	    && read ans && [ $${ans:-'N'} = 'y' ]
	platformio project init --ide qtcreator

# forward any other target (clean, build, etc.) to pio run
%:
	platformio -c qtcreator run --target $*
```
Now, in addittion to the build and clean steps setted autumatically when creating the project, it is only necessary to define that the Deployment will be with make and define _deploy_ as an argument
Add Deploy Step -> make

ksnip_20220523-135729

It is no longer necessary to define any executable directly in QtCreator. It shows a warning about it, but I can live with it.

In the case of bash, it doesn't work even if I do it from terminal

`bash -c platformio device monitor`

It shows me the platformio help, as if I hadn't invoked it correclty.

It would be interesting to see how to do it with bash, which comes by default in most popular distros.

Thanks again for your help!

dewhisna:
@JoseOlin -- it's probably not doable from just bash alone or at least not easily.  But I figured it was worth a quick test to try it.  The ioctl error was coming from the termios library that apparently platformio is using during that operation and was complaining because the stdio channels passed in from Creator weren't "terminal" devices.  I haven't dug into what ioctl things it's sending to the terminal, but it's probably related to buffering or echos or something similar.  To work with bash, you'd have to have it connect to different stdio streams when it calls platformio instead of it passing the ones through from Creator, ones that support terminal ioctl.  Not totally impossible, but certainly a lot more involved than just using xterm.

JoseOlin:
Oh, I see...

Well xterm is more than enough for me as I was just looking for a way to not have to manually configure each new project in Creator.

Thanks for all your support!

belm0:
> `bash -c platformio device monitor`
> 
> It shows me the platformio help, as if I hadn't invoked it correclty.

I believe it should be `bash -c "platformio device monitor"`

belm0:
@JoseOlin with your setup, if you hit "Run" again, is the xterm automatically killed and everything works fine?

> xterm -hold -e platformio device monitor

We could consider adding it to the default Makefile, though xterm is not really cross-platform (I can run it on mac, but only by installing xquartz ad running a server).

Also I'm wondering if `-hold` is really needed, since the only reason `device monitor` would exit is if you disconnect the device or explicitly close it by command shortcut.

PR #4267, opened 2022-05-15
qtcreator: add project-update makefile target

belm0:


belm0:
Actually it doesn't work consistently.  Especially handling of platformio.config is very mysterious.  It seems to remember old defines that were removed from platformio.ini.

belm0:
> Especially handling of platformio.config is very mysterious. It seems to remember old defines that were removed from platformio.ini.

It only works if I delete `.pio/` first.  It has something to do with `idedata.json`, there is an unwanted feedback loop.

belm0:
@ivankravets I think this PR was closed unintentionally, since we still want the `project-update` target added

ivankravets:
Oh, yes, sorry. Please remove that part that removes .pio internals.

PR #4297, opened 2022-05-31
Append files from any folder in src_dir starting with ino_ too

arendst:
This allows for moving files from the root to sub-folders like moving all sensor drivers to one folder called ino_xsns and other drivers to a folder called ino_xdrv. The naming is still important as the InoToCpp conversion needs files in alphabetical order.

I've tested this with Tasmota andmoving the drivers into 6 different folders.

### DIFF
--- a/platformio/builder/tools/pioino.py
+++ b/platformio/builder/tools/pioino.py
@@ -230,6 +230,10 @@ def ConvertInoToCpp(env):
     ino_nodes = env.Glob(os.path.join(src_dir, "*.ino")) + env.Glob(
         os.path.join(src_dir, "*.pde")
     )
+    # append files from any folder in src_dir starting with ino_ too
+    ino_nodes = ino_nodes + env.Glob(os.path.join(src_dir, "ino_*", "*.ino")) + env.Glob(
+        os.path.join(src_dir, "ino_*", "*.pde")
+    )
     if not ino_nodes:
         return
     c = InoToCPPConverter(env)


arendst:
@ivankravets Just to let you know.

arendst:
image

ivankravets:
I believe in a day when we will finally drop support for INO files. You can't imagine how many days were lost due to the INO files. I even don't know who invented this shit. People don't understand that INO != CPP. We are so tired to teach everyone. We even create a guide on how to convert INO -> CPP => https://docs.platformio.org/en/latest/faq/ino-to-cpp.html

So, I'm going to close this PR and will think about adding a warning in the next release to the projects that use outdated legacy INO files.

If someone needs INO, then the right way is to use Arduino IDE. We don't plan to invest more in this bullshit format.

P.S: Thanks for using PlatformIO!

Jason2866:
@ivankravets I can understand the frustration of people who do not understand that INO != CPP.
However, Arduino (INO) has brought many people (who had nothing to do with programming) to this great hobby. Not only that many great projects are based on Arduino and the myriad of libraries that make it possible not to have to reinvent the wheel every time. For professional projects, Arduino IDE is not an option. You know this best!
Tasmota would not be possible with Arduino IDE. It would certainly also be a shame to lose the many users of espressif MCUs. I think a significant number of Platformio users are users of these MCUs. Most of them use Arduino (probably because of the available libraries).
Maybe it is worth reconsidering to support INO in the future. Knowing that this format has considerable downsides. But which programming language does not have this?
My 2 cents
Jason

ivankravets:
Does Tasmota support Arduino IDE too? I'll think about how to provide an API for you where you can modify INO->CPP for your needs.

What I mean, is that we don't plan to invest more in the support for INO. The current implementation reproduces the behavior of classic Arduino IDE. Extending or modifying our implementation could lead to new issues.

I'll back soon here.

Jason2866:
No, Tasmota is impossible to build with Arduino IDE. Tasmota uses and needs features only available with PlatformIO.
If you look at the Github Page you see
Bildschirmfoto 2022-05-31 um 21 11 30

ivankravets:
Thanks, got you. What is the reason that you still use INO?

Jason2866:
Avoiding the pain to migrate ;-).  Tasmota started as small Arduino IDE sketch and has grown over the years.
We considered some time ago. Looking at the work to do we discarded.
It is a free time project of one person with some contributors (doing in free time too).

arendst:
> What is the reason that you still use INO?

As said by @Jason2866 we started off using Arduino IDE. Currently the project uses over 300 files successfully merged by your great ``pioino.py`` script into one big .cpp file of over 174500 (!!!) lines of code. By using ``.ino`` files we do not have to bother about prototypes and where to store them; they are inserted at the correct location in the large file.

To manage these number of files I'm just proposing a transparent way of moving some files to subdirectories still keeping in mind the ``.ino`` philosophy regarding file naming (which controls the order of the files within the final ``.cpp`` file.

The proposed solution has no impact for "normal" .ino users using the project source folder as the main code area. It allows Tasmota to move almost 250 files into 6  subdirectories.

ivankravets:
I've just added a new API to the build env. You can implement your own methods that return INO nodes. Example of use:

```ini
[env:myenv]
extra_script = pre:script.py
```

`script.py`

```python

import glob
import os

Import("env")

def FindInoNodes(env):
    src_dir = glob.escape(env.subst("$PROJECT_SRC_DIR"))
    return env.Glob(os.path.join(src_dir, "*.ino")) + env.Glob(
        os.path.join(src_dir, "*.pde")
    )

env.AddMethod(FindInoNodes)

```

Does it work for you?

It will be a part of PIO Core 6.0.2. We plan to release it today/tomorrow.

arendst:
Thx. Will do some testing now.

Issue #4332, opened 2022-06-26
Builds using atmelsam framework versions 3.9.0 to 3.5.0 fail in PIO Core 6.0.2


ollie1400:
What kind of issue is this?

- [ ] **Question**.
      This issue tracker is not the place for questions. If you want to ask how to do something,
      or to understand why something isn't working the way you expect it to, 
      use [Community Forums](https://community.platformio.org) or [Premium Support](https://platformio.org/support)

- [ ] **PlatformIO IDE**.
      All issues related to PlatformIO IDE should be reported to appropriate repository:
      [PlatformIO IDE for Atom](https://github.com/platformio/platformio-atom-ide/issues) or
      [PlatformIO IDE for VSCode](https://github.com/platformio/platformio-vscode-ide/issues)

- [ ] **Development Platform or Board**.
      All issues (building, uploading, adding new boards, etc.) related to PlatformIO development platforms
      should be reported to appropriate repository related to your hardware
      https://github.com/topics/platformio-platform

- [ ] **Feature Request**.
      Start by telling us what problem you’re trying to solve. Often a solution
      already exists! Don’t send pull requests to implement new features without first getting our
      support. Sometimes we leave features out on purpose to keep the project small.

- [x] **PlatformIO Core**.
      If you’ve found a bug, please provide an information below.

*You can erase any parts of this template not applicable to your Issue.*

------------------------------------------------------------------

### Configuration

**Operating system**:

Microsoft Windows 10 Pro for Workstations
10.0.19043 Build 19043

**PlatformIO Version** (`platformio --version`):

PlatformIO Core, version 6.0.2

### Description of problem

Projects that use `atmelsam` framework versions `3.9.0` to `3.5.0` fail to build in version `6.0.2` of PIO, but build correctly on version `5.1.0`.

This seems to be because
```
from platformio.device.list import list_serial_ports as get_serialports
```
in `util.py` was changed to
```
from platformio.device.list import list_serial_ports as get_serial_ports
```
but `main.py` in `atmelsam@3.6.1` (and others, see `Additional info` below) is looking for 
```
from platformio.util import get_serialports
```

#### Steps to Reproduce

1. Create a project using board `Arduino MKRZERO` and framework `Arduino`
2. Build the project with `pio run`

### Actual Results

Build fails:

```
> pio run

Processing mkrzero (platform: atmelsam@3.6.1; board: mkrzero; framework: arduino)
[...]

PR #4333, opened 2022-06-26
Export twice to maintain backward compattability

ollie1400:
Export function also using old name to be backward compatible.  Solve https://github.com/platformio/platformio-core/issues/4332


ivankravets:
Thanks for the PR!

Issue #4391, opened 2022-08-12
Unescaped File Path Injected into Python String Constant in test_build_unflags


COTSWORKS-MackeyK:
In the unit test `test_build_unflags`:

https://github.com/platformio/platformio-core/blob/8814f4e92d37822abbcbdc973fb2f9f2ac40b7a4/tests/commands/test_run.py#L123

A python script is generated by using the `%` Python operator to format a string:

https://github.com/platformio/platformio-core/blob/8814f4e92d37822abbcbdc973fb2f9f2ac40b7a4/tests/commands/test_run.py#L143-L153

The argument that is injected is a string representation of an absolute file path:

https://github.com/platformio/platformio-core/blob/8814f4e92d37822abbcbdc973fb2f9f2ac40b7a4/tests/commands/test_run.py#L153

The path is injected between double quotes to act as a string constant in the resulting Python file:

https://github.com/platformio/platformio-core/blob/8814f4e92d37822abbcbdc973fb2f9f2ac40b7a4/tests/commands/test_run.py#L145

On Windows, the file path separator is the backslash, `\`.

This can created a syntax error in the generated Python file if the backslash separator precedes a character that creates an escape sequence, such as `\t`.

COTSWORKS-MackeyK:
My first attempt at resolving this issue:

https://github.com/COTSWORKS-MackeyK/platformio-core/commit/49ede3e57c99ad37b9614f7f96dcc3191b926bdb

COTSWORKS-MackeyK:
Demonstration:
* Dell Inc., XPS 15 9500, x64
* Microsoft Windows 11 Pro 10.0.22000 Build 22000
* Git 2.37.1
* Python 3.7.9
* `tox` 3.25.1
* PowerShell 7.2.5

Procedure:
1. `git clone https://github.com/platformio/platformio-core.git`
2. `cd platformio-core`
3. `git submodule init`
4. `git submodule update`
5. `tox -e py37`
6. `.\.tox\py37\Scripts\activate.ps1`
7. `py.test -n 1 tests --ignore tests/test_examples.py -k test_build_unflags`

Result:
```
$ (py37) > py.test -n 1 tests --ignore tests/test_examples.py -k test_build_unflags
============================================================================================= test session starts ==============================================================================================
platform win32 -- Python 3.7.9, pytest-7.1.2, pluggy-1.0.0
rootdir: C:\*********\platformio-core, configfile: tox.ini
[...]

PR #4394, opened 2022-08-15
FIX  Unescaped File Path

COTSWORKS-MackeyK:
Fix Issue #4391 in platformio/platformio-core

https://github.com/platformio/platformio-core/issues/4391

Use Python string.replace(find, replace) method to escape backslash
`\` characters with a preceding backslash.

`C:\A\path\that\is.unescaped`

becomes

`C:\\A\\path\\that\\is.unescaped`

Preventing the '\t' from being interpreted as a special character.

It also prevents Python syntax errors from unrecognized ASCII encoding sequences.


ivankravets:
Let's discuss it in https://github.com/platformio/platformio-core/issues/4391

PR #4408, opened 2022-09-03
Added CLion IDE CTest support

hacker-cb:
This is very useful to run tests directly from the CLion IDE.

### DIFF
--- a/platformio/project/integration/generator.py
+++ b/platformio/project/integration/generator.py
@@ -22,6 +22,7 @@ import bottle
 from platformio import fs, util
 from platformio.proc import where_is_program
 from platformio.project.helpers import load_build_metadata
+from platformio.test.helpers import list_test_suites
 
 
 class ProjectGenerator:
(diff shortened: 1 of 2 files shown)


hacker-cb:
Hi @ivankravets. Could you check this PR please?

hacker-cb:
Actually I solved this by creating custom [CMakeListsUser.txt](https://gist.github.com/hacker-cb/8fe5e187557e230e374af3d3077effb1). No need to modify PlatformIO core.

hacker-cb:
> Actually I solved this by creating custom [CMakeListsUser.txt](https://gist.github.com/hacker-cb/8fe5e187557e230e374af3d3077effb1). No need to modify PlatformIO core.

Project was moved to https://github.com/hacker-cb/platformio-cmake

MacDada:
> Project was moved to https://github.com/hacker-cb/platformio-cmake

Nice, I'm definitely gonna check it out.

You closes the issue though, wouldn't that make sense for this to be included in PIO? I guess I'd have to be lucky to find your repo with the solution – it would be nice if PIO and CLion worked together OOTB.

hacker-cb:
> > Project was moved to https://github.com/hacker-cb/platformio-cmake
> 
> Nice, I'm definitely gonna check it out.
> 
> You closes the issue though, wouldn't that make sense for this to be included in PIO? I guess I'd have to be lucky to find your repo with the solution – it would be nice if PIO and CLion worked together OOTB.

Actually it is very simple to add any additional logic using `CMakeListsUser.txt`.
 My additions is just part of logic which can be useful by PlatdormIO users who use Clion or other CMake based IDEs.
I agreed with developers that no need to include custom logic into the core. It can be done with custom CMake files.

PR #4427, opened 2022-10-01
Add udev Entry for Wio Terminal

mymro:
For me at least those entries are necessary. Even after adding my user to the relevant groups it refused to connect to the device.


ivankravets:
Thanks for the PR!

Issue #4432, opened 2022-10-03
Correct use of corporate proxy


LioRei:
What kind of issue is this?

- [x] **PlatformIO Core**.
      If you’ve found a bug, please provide an information below.

*You can erase any parts of this template not applicable to your Issue.*

------------------------------------------------------------------

### Configuration

**Operating system**: Windows

**PlatformIO Version** (`platformio --version`): PlatformIO Core, version 6.1.4

### Description of problem

Platformio is not able to correctly install behind a corporate proxy. HTTPClient and FileDownloader isn't reading env variable to see if there is a proxy or not so the download of platformio/contrib-piohome is not possible.

#### Steps to Reproduce

1. have a computer behing a corporate proxy
2. install platformio
3. try to do a "pio home"

### Actual Results

Error in download of the package

### Expected Results

No error 

### Additional info
I've modified the python script on my computer and I will link a draft PR with my change

LioRei:
draft PR: https://github.com/platformio/platformio-core/issues/4432

ivankravets:
Could you re-test with `pio upgrade --dev`? See newly added docs on how to configure the proxy https://docs.platformio.org/en/latest/core/installation/proxy-configuration.html

PR #4433, opened 2022-10-03
Draft for corporate proxy

LioRei:
it's a draft (and ugly solution) for being able to install package behind a corporate proxy.

Optimal solution will be to take VS code configuration to do this but.... I'm not a python dev and I really don't know how to do this

### DIFF
--- a/platformio/package/download.py
+++ b/platformio/package/download.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import io
+import os
 from email.utils import parsedate
 from os.path import getsize, join
 from time import mktime
@@ -24,6 +25,7 @@ from platformio import __default_requests_timeout__, app, fs
 from platformio.compat import is_terminal
 from platformio.package.exception import PackageException
 
+proxy = {"http":os.getenv('HTTP_PROXY'), "https":os.getenv('HTTPS_PROXY')}
 
 class FileDownloader:
     def __init__(self, url, dest_dir=None):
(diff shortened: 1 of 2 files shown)


platformio/package/download.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, url, dest_dir=None):
         self._request = requests.get(
             url,
             stream=True,
+            verify=False,
```
maxgerhardt, line 35: This weakens security for *everyone* not using a proxy, opening them up to man-in-the-middle attacks, supply chain poisoning etc. Can we only ocnditionally do this if another option / environment variable is enabled, such as "Disable strict SSL verifcation"? Same for the line above.

LioRei: Of course, like this it's a big security concern ;)

We could base this on the environment variable [NODE_TLS_REJECT_UNAUTHORIZED](https://nodejs.org/api/cli.html#node_tls_reject_unauthorizedvalue) ?

ivankravets: Yes, we will provide environment variable to disable TLS Certificates verification

mcspr:
afaik 'requests' automatically picks up `*_PROXY` environment variable, so it is not necessary to inject ourselves?
https://requests.readthedocs.io/en/latest/user/advanced/#proxies
https://docs.python.org/3/library/urllib.request.html#urllib.request.getproxies (with some exceptions)
https://github.com/psf/requests/blob/v2.28.1/requests/utils.py (/proxies/)

LioRei:
that's what the doc says but that's not what happens in real life... I tried again and I confirm that even with the env variable, the request does not take care of the proxy

ivankravets:
Resolved in https://github.com/platformio/platformio-core/issues/4432

PR #4450, opened 2022-11-02
Add list_logical_devices to public

not7cd:
This would help platform developers ensure compatibility between versions. 
In short, I ran for hours after a changed import. 
I think this commit introduced this issue: https://github.com/platformio/platformio-core/commit/a6e61a7a5a24cf73d2866b9817e8b9a7e99852d0

This would save others from that: https://github.com/maxgerhardt/platform-raspberrypi/issues/15

### DIFF
--- a/platformio/public.py
+++ b/platformio/public.py
@@ -14,7 +14,7 @@
 
 # pylint: disable=unused-import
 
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR!

PR #4479, opened 2022-11-28
Legacy docs urls changed to the current structure

MacDada:
I just opened those URLs to see where they point right now.

* https://github.com/platformio/platformio-docs/issues/275#issuecomment-1329510718
* https://github.com/platformio/platformio-docs/pull/276


MacDada:
Should I rebase/update the PR?

ivankravets:
There is no need to update all URLs. The `/page/` is a valid mount point according to the Sphynx and RTD. There are redirects from the "entry point" to the valid language-based page.

PR #4481, opened 2022-11-30
add udev rule for AIR32F103 board

mjysci:
It's a low cost dev board that can use as DAPLink debugger. [info](https://wiki.luatos.com/chips/air32f103/board.html)  
This PR fixed `Error: unable to open CMSIS-DAP device 0xd28:0x204` while running upload task.


ivankravets:
Thanks for the PR!

PR #4494, opened 2022-12-19
Dont error when no `upload port` found with Platform `darwin`

Jason2866:
Currently compile breaks when no `upload port` is found when using "autodetect"

We have this error when building Tasmota with GH actions CI, when trying to use `autodetect `since there are no ports available.

@ivankravets Would be great if this fix or a better one from you ;-), would be in next release.

Thx for your great work!

Greets from Tasmota crew. Stay safe.

### DIFF
--- a/platformio/builder/tools/pioupload.py
+++ b/platformio/builder/tools/pioupload.py
@@ -24,6 +24,7 @@ from SCons.Script import ARGUMENTS  # pylint: disable=import-error
 from serial import Serial, SerialException
 
 from platformio import exception, fs
+from platformio.util import get_systype
 from platformio.device.finder import SerialPortFinder, find_mbed_disk, is_pattern_port
 from platformio.device.list.util import list_serial_ports
 from platformio.proc import exec_command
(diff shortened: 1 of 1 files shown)


ivankravets:
Hm... That part of the code is called only on the "upload" target. Could you provide a simple project to reproduce this issue?

Jason2866:
Simple not really, the CI branch (with the patch in the GH CI) is [here:](https://github.com/Jason2866/Tasmota/tree/uploadport) The used [CI yaml](https://github.com/Jason2866/Tasmota/blob/uploadport/.github/workflows/build_all_the_things.yml)

Jason2866:
Will do, need a bit of time for.

Jason2866:
Can not reproduce. Setup is [here](https://github.com/Jason2866/Pio_Mac_upload_port)
Why the orig. setup is failing???

Jason2866:
Found the reason it is [here](https://github.com/arendst/Tasmota/blob/development/pio-tools/download_fs.py#L32)
Maybe you can implement a Filesystem Downloader, so this custom Script is not needed anymore?

PR #4498, opened 2022-12-25
Allow change of project_name variable

pedro-vo:
This commit intends to introduce new feature of manally configurable project name which is exported to various IDE configurations. New configuration option in platformio.ini in section [platformio] is introduced: project_name. When this option is set, the current (default) value of project_name which is get from project directory basename is replaced by the configured value.


ivankravets:
Please re-test with `pio upgrade --dev`. The new option is `name`.

PR #4521, opened 2023-01-16
changed bmp gdb init

reimerp:
# Blackmagic Probe improvements

I needed SRST under reset feature, this command has to be issued before scan, but this is currently impossible with current variables.

By changing the gdb init a bit this or even target power is possible

See: "Embedded Debugging with the Black Magic Probe", pp. 28

```
upload_protocol = blackmagic
upload_port = /dev/ttyBmpGdb
debug_tool = blackmagic
debug_port = ${this.upload_port} connect_srst
monitor_port = /dev/ttyBmpTarg
monitor_speed = 115200

```

debug_port = port [monitor_cmd]

debug_port can be /dev/ttyACM0 (/dev/ttyBmpGdb if you use the udev rules) or a Windows COM port, I use `${this.upload_port}` to use the same.

An optional param can be used to enable eg:

- tpwr (Target Power)
- connect_srst (Assert SRST during connect)

Currently it's not possibe to enable both, but this can easily be changed (iterate over args)

### DIFF
--- a/platformio/debug/config/blackmagic.MD
+++ b/platformio/debug/config/blackmagic.MD
@@ -1,0 +1,28 @@
+# Blackmagic Probe improvements
+
+I needed SRST under reset feature, this command has to be issued before scan, but this is currently impossible with current variables.
+
+By changing the gdb init a bit this or even target power is possible
+
+See: "Embedded Debugging with the Black Magic Probe", pp. 28
+
+```
+upload_protocol = blackmagic
+upload_port = /dev/ttyBmpGdb
+debug_tool = blackmagic
+debug_port = ${this.upload_port} connect_srst
+monitor_port = /dev/ttyBmpTarg
+monitor_speed = 115200
+
+```
+
+debug_port = port [monitor_cmd]
+
+debug_port can be /dev/ttyACM0 (/dev/ttyBmpGdb if you use the udev rules) or a Windows COM port, I use `${this.upload_port}` to use the same.
+
+An optional param can be used to enable eg:
+
+- tpwr (Target Power)
+- connect_srst (Assert SRST during connect)
+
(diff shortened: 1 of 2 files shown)


ivankravets:
@esden, could you review this PR?

platformio/debug/config/blackmagic.MD
```diff
@@ -0,0 +1,28 @@
...
+debug_tool = blackmagic
+debug_port = ${this.upload_port} connect_srst
+monitor_port = /dev/ttyBmpTarg
+monitor_speed = 115200
+
+```
+
+debug_port = port [monitor_cmd]
+
+debug_port can be /dev/ttyACM0 (/dev/ttyBmpGdb if you use the udev rules) or a Windows COM port, I use `${this.upload_port}` to use the same.
```
dragonmux, line 21: Small nitpick, /dev/ttyACM0 is unreliable in name - the kernel gives us no guarantees that it didn't bring BMP up as some other name instead, hence the project udev rules. Second note, if a user wishes to connect to a specific BMP by serial number, they can update their udev rules and use /dev/ttyBmpGdb<serial>. The udev rules and serial number stuff is stable through releases

dragonmux (approved):
So, not Esden, obviously, but rather the lead dev for the Black Magic Debug project. This change looks good to us with some notes (below) for general improvements that could be made that future-proof this a bit and bring the PlatformIO notes on BMD slightly more into line with the upstream project and the direction being taken in the future.

platformio/debug/config/blackmagic.py
```diff
@@ -20,12 +20,25 @@
 class BlackmagicDebugConfig(DebugConfigBase):
```
RawRaw111, line 20: 20

PR #4598, opened 2023-04-13
bugfix: get_original_version bug for versions with a git sha

shlomozippel:


ivankravets:
Could you provide an example?

PR #4625, opened 2023-05-13
`__eq__`s check for the class of `other`s

MacDada:
Fixes errors like…

```
AttributeError: 'Version' object has no attribute 'owner':
  File "/usr/local/Cellar/platformio/6.1.6/libexec/lib/python3.11/site-packages/platformio/package/meta.py", line 188
```

…when trying to compare the objects to other objects that are not of the same class.

---

I came up with the issue when trying to debug my [PIO's extra script](https://docs.platformio.org/en/latest/projectconf/sections/env/options/advanced/extra_scripts.html) with `var_dump`:

```python
Import("env")
env.Execute("$PYTHONEXE -m pip install 'var_dump @ git+https://github.com/sha256/python-var-dump@972d05bff13fd57366f56bd36b70e1ab15d6173d' --force-reinstall")
from var_dump import var_dump
var_dump(env.PioPlatform())
```

### DIFF
--- a/platformio/package/commands/outdated.py
+++ b/platformio/package/commands/outdated.py
@@ -36,7 +36,7 @@ class OutdatedCandidate:
             self.envs = [self.envs]
 
     def __eq__(self, other):
-        return all(
+        return isinstance(other, OutdatedCandidate) and all(
             [
                 self.pm.package_dir == other.pm.package_dir,
                 self.pkg == other.pkg,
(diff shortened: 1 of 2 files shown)


ivankravets:
Could you point to the place in the PIO Core where the invalid type is passed to the "compare" methods?

ivankravets:
Please reopen if you still need help

PR #4661, opened 2023-06-15
Fix CMakeLists.txt.tpl to use local compiler

AdelCraftZ:
The CMakeListsPrivate.txt is included after defining the project (with the "project" command). It should be the other way around since the C and C++ compilers are defined in CMakeListsPrivate.txt. The compiler definition always goes before the project definition, otherwise it is ignored by CMake. Since it is not the case here, CLion will tell CMake to try detecting the system compiler, which is probably not the one being used by the project. This may works if it the system compiler is a GNU compiler, but will not work otherwise (eg. with MSVC installed and set as the default in CLion). The solution is simply to include CMakeListsPrivate.txt before defining the project.


MacDada:
I don't know if this is relevant, but you might wanna hold your horses, as they just finished: https://youtrack.jetbrains.com/issue/CPP-22328

PR #4674, opened 2023-06-21
ProjectOptionValueError: displays the config option description

MacDada:
It gives a bit more context of the problem for the user.

BEFORE:

`Error: Invalid value: 'invalid_debug_mode' is not one of 'always', 'modified', 'manual'. for option `debug_load_mode` in section [env:nodemcu]`

AFTER:

`Error: Invalid value: 'invalid_debug_mode' is not one of 'always', 'modified', 'manual'. for option `debug_load_mode` (Allows one to control when PlatformIO should load debugging firmware to the end target) in section [env:nodemcu]`


ivankravets:
Ah, I see, you updated exception as well.

Issue #2872, opened 2019-08-05
Use XDG_CONFIG_HOME for .platformio directory


microbug:
### Configuration

**Operating system**: macOS 10.14

**PlatformIO Version** (`platformio --version`): 4.0.1b3

### Description of problem
The `~/.platformio` directory contains important config data including the user's custom boards. It's currently one of many hidden directories in my home folder. For most Unix-based systems these directories conventionally go in `$XDG_CONFIG_HOME`, which is usually `~/.local/share`. This makes things neater for the user.

On Windows and macOS the variables `XDG_CONFIG_HOME` usually aren't defined, so you can use `C:\Users\x\AppData\Roaming` and `~/.local/share` respectively (i.e., `~/.local/share` does exist on macOS but isn't usually defined as `XDG_CONFIG_HOME`).

#### Steps to Reproduce

1. Run `pio init`
2. Observe new `.platformio` directory in `$HOME`
3.

peterhoeg:
```$XDG_CONFIG_HOME``` is usually ```~/.config``` while ```$XDG_DATA_HOME``` is ```~/.local/share```.

ivankravets:
Yes, we keep the same location through all systems. You can override it with http://docs.platformio.org/en/latest/envvars.html#envvar-PLATFORMIO_CORE_DIR

lhanson:
I think that allowing the user to override PIO's defaults is besides the point; that's a whole lot of directories which a user is forced to remap. The goal of the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is to free the user from the burden of having to manually manage each piece of software as a unique snowflake directory tree dumped somewhere random in their $HOME directory.

For example, what subdirectories of `~/.platformio` are configuration which needs to be backed up and can be synced between development machines? What directories possibly contain credentials which should never be backed up? Which are runtime data which is necessary to have on hand? Which are cache directories and can safely be discarded if disk space is low? The XDG spec is a way to avoid the user having to figure out the answers to those questions for each piece of software we install. If PIO respected the preferences I've already expressed (and shouldn't have to say differently for every program I install), I'd know that `~/.config/platformio/*` is configuration that I should back up, `~/.cache/platformio` is safe to delete and can be recreated as needed, etc.

Please reconsider opening this issue. Complying with the XDG spec brings a lot of benefits and loses nothing. It's trivial to remain backwards compatible to support existing installations.

[...]

PR #4680, opened 2023-06-26
Comply with XDG specification

Jappie3:
Modified `get_default_core_dir()` so that it respects the XDG specification and does not clutter people's home directories with `.platformio` folders. See issue #2872.

I left Windows alone since they don't really do XDG - I believe they have their own conventions, e.g. `%APPDATA%` and stuff like that. Platformio also doesn't adhere these, but I'll leave that to someone else to figure out.

Besides that, I also had to modify `test_config.py` so that it would not freak out because of the changes in `options.py`.

More info about the XDG specification:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
https://xdgbasedirectoryspecification.com/

### DIFF
--- a/platformio/project/options.py
+++ b/platformio/project/options.py
@@ -118,7 +118,11 @@ def validate_dir(path):
 
 
 def get_default_core_dir():
-    path = os.path.join(fs.expanduser("~"), ".platformio")
+    xdg_config_home = os.environ.get("XDG_CONFIG_HOME")
+    if xdg_config_home:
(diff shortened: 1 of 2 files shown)


Jappie3:
Thinking about this a little bit more, it doesn't really make sense to dump everything XDG_CONFIG_HOME... maybe XDG_DATA_HOME is better suited? or should the files be divided over config & data (that would complicate things)

PR #4773, opened 2023-10-29
Add STM32 virtual COM port to `udev.rules`

jake-is-ESD-protected:
In case of interfacing an STM32 chip as USB-CDC device. This enabled me to treat the USB-port of the Electrosmith Daisy Seed as a normal `/dev/ttyACMx` serial interface. Other use-cases are not known to me, you guys decide if that is a worthy addition.


ivankravets:
Thanks for the PR!

PR #4786, opened 2023-11-14
Remove unused build output buffer

### LINKED ISSUE #4783: Verbose build takes 75x as long as regular build
### Configuration

**Operating system**: macOS 14.1.1 (arm64)

**PlatformIO Version**: PlatformIO Core, version 6.1.11

### Description of problem

Building my project with `--verbose` significantly increases the time to build.

#### Steps to Reproduce

I have a very minimal project based on `espidf`:

<details>
<summary>platformio.ini</summary>

```
[env:test]
platform = espressif32 @ 5.4.0
framework = espidf

# Settings for ESP32-S3-WROOM-1U-N16
board = esp32-s3-devkitc-1
board_build.flash_mode = dio
board_build.partitions = partitions.csv
board_upload.flash_size = 16MB
board_upload.before_reset = no_reset
board_upload.after_reset = no_reset
```
</details>

Building regularly:

```
$ pio run -t clean && pio run
...
=== [SUCCESS] Took 22.97 seconds ===
```

Building verbosely:

```
$ pio run -t clean && pio run -v
...
=== [SUCCESS] Took 1722.01 seconds ===
```

rgov:
In verbose mode, output from the build command was constantly being appended to an unused buffer, character by character. Even a simple project using ESP-IDF would generate over 5 million append operations, which is quite slow because the backing memory for this buffer has to be constantly reallocated.

The actual buffer that's used is the local `line` variable. An alternative fix would be to remove `line` and use `self._buffer` instead, so that the `get_buffer()` method will return the buffer contents.

| | Time to compile |
| --- | --- |
| Baseline (non-verbose) | 22.97 seconds |
| Before patch (verbose) | **1722.01 seconds** |
| After patch (verbose) | 24.23 seconds |

Fixes #4783.

### DIFF
--- a/platformio/proc.py
+++ b/platformio/proc.py
@@ -69,6 +69,8 @@ class BuildAsyncPipe(AsyncPipeBase):
         print_immediately = False
 
         for char in iter(lambda: self._pipe_reader.read(1), ""):
+            # self._buffer += char
+            
             if line and char.strip() and line[-3:] == (char * 3):
                 print_immediately = True
 
--- a/platformio/proc.py
+++ b/platformio/proc.py
@@ -70,7 +70,7 @@ class BuildAsyncPipe(AsyncPipeBase):
 
         for char in iter(lambda: self._pipe_reader.read(1), ""):
             # self._buffer += char
-            
+
             if line and char.strip() and line[-3:] == (char * 3):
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR!

PR #4830, opened 2024-01-12
Allow Starlette 0.35.1

musicinmybrain:
I need to loosen the dependency bound for the package in the development version of Fedora Linux, so I’m offering the change as a PR.

I had a hard time validating this, as some tests were failing in my environment before this commit, but I don’t expect there should be anything disruptive here.

https://github.com/encode/starlette/releases

### DIFF
--- a/platformio/pipdeps.py
+++ b/platformio/pipdeps.py
@@ -35,7 +35,7 @@ def get_pip_dependencies():
     home = [
         # PIO Home requirements
         "ajsonrpc == 1.2.*",
-        "starlette >=0.19, <0.35",
+        "starlette >=0.19, <0.36",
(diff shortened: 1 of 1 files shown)


ivankravets:
Hi,  

thanks for the PR. Have you seen https://github.com/platformio/platformio-core/blob/ffacd173873cebf0f980585d4da018c286a99b42/platformio/pipdeps.py#L38?

musicinmybrain:
> Hi,
> 
> thanks for the PR. Have you seen
> 
> https://github.com/platformio/platformio-core/blob/ffacd173873cebf0f980585d4da018c286a99b42/platformio/pipdeps.py#L38
> ?

No, I missed that. Perhaps I was working with a slightly stale checkout. I’ll try to get the package updated to the latest release instead of patching the dependency. Thanks!

PR #4870, opened 2024-03-03
Fix symlink package spec validation

willmmiles:
While developing a library package, I tried to use a `symlink://` specification in my test application to point to my library's development checkout.  Unfortunately this seemed to cause a number of issues:
- The symlinked package did not appear in `platformio pkg list`
- Every build would "re-install" the symlink package
- As a result of the "reinstallation", the PlatformIO VSCode plugin would try to re-configure itself mid-build, causing a number of failures as it deleted some of the working build folders

I traced this to the installed symlink package not matching its own pkg spec in `package/manager/base.py`.   It looks like it just needed the same `abspath()` check as a `file://` specification.


ivankravets:
Thanks for the PR!

PR #4874, opened 2024-03-12
paths fix for check tool

adrianstephens:
My src_dir and include_dir are not in the same directory as the platformio.ini, which prevented the check tool from working.
I found out why, and this is my fix!


ivankravets:
Thanks for the PR!  Please re-test with `pio upgrade --dev`.

PR #4880, opened 2024-03-23
Update 99-platformio-udev.rules

soiamsoNG:
Fix typo on rule for Espressif USB JTAG/serial debug unit


ivankravets:
Thanks for the PR!

PR #4886, opened 2024-03-26
Improve memory exhaustion warnings/errors

Sophist-UK:
Firmware typically pre-allocates RAM memory where it can, and then remaining memory is used for dynamic allocation and the stack.

Currently excessive RAM usage is only a warning and not an error, which means that you are not alerted by a build failure but instead need to scan the PIO RUN output to see whether RAM usage is too high for the build. But in any case, typically firmware requires a certain amount of RAM to be available after the pre-allocation that is calculated by PIO, and you need to alerted when the remaining memory falls below this required free-memory amount.

This PR does two things:

1. RAM pre-allocation greater than available memory is now treated as an error and not a warning.
2. It introduces a RAMFREE option that can be used to define a minimum amount of memory required to remain free (for dynamic allocation and stack usage) after pre-allocation, with an error if pre-allocation results in less than this amount free.

### DIFF
--- a/platformio/builder/tools/pioupload.py
+++ b/platformio/builder/tools/pioupload.py
@@ -218,16 +218,20 @@ def CheckUploadSize(_, target, source, env):
     if int(ARGUMENTS.get("PIOVERBOSE", 0)):
         print(output)
 
+    memory_exceeded = False
     if data_max_size and data_size > data_max_size:
+        memory_exceeded = True
         sys.stderr.write(
             "Warning! The data size (%d bytes) is greater "
             "than maximum allowed (%s bytes)\n" % (data_size, data_max_size)
         )
     if program_size > program_max_size:
+        memory_exceeded = True
         sys.stderr.write(
             "Error: The program size (%d bytes) is greater "
             "than maximum allowed (%s bytes)\n" % (program_size, program_max_size)
         )
+    if memory_exceeded:
         env.Exit(1)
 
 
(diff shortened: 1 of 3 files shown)


Sophist-UK:
P.S. This simple PR has been coded on Github and I have not tested it. Sorry - but I don't have the skills or time to create a local dev environment and test it.

ivankravets:
> ram_free

You can override this value via `board_upload_maximum_ram_size` (RAM) and `board_upload_maximum_size` (Flash)

Sophist-UK:
@ivankravets said:
> You can override this value via `board_upload_maximum_ram_size` (RAM)

1. I am not sure how to provide these.
2. Since I have automated builds for a couple of different boards with different amounts of memory from the same source code, if I am specifying maximum ram size I need to be able to set this dynamicaly - any advice here would also be useful.

(P.S. Flash is all or nothing - it is RAM which is more nuanced.)

Sophist-UK:
> In the platformio.ini [env] section?

I tried this and it didn't work.

ivankravets:
Could you provide your INI file on how you declared these lines?

Sophist-UK:
[platformio.ini.txt](https://github.com/platformio/platformio-core/files/14760844/platformio.ini.txt)

ivankravets:
Sorry, this is a my typo. See updated https://github.com/platformio/platformio-core/pull/4886#issuecomment-2020190694 

The JSON path must be split with `.` dot. https://github.com/platformio/platform-atmelavr/blob/develop/boards/megaatmega2560.json#L33

Sophist-UK:
@ivankravets I tried the dotted parameter name in the same place in the ini file and it didn't reduce the target ram size from 8k to 6.5K as it should have. Sorry.

Issue #4769, opened 2023-10-26
Feature request: Make COMPILATIONDB_INCLUDE_TOOLCHAIN as compiledb flag


jayadamsmorgan:
I am working with a lot of projects and my main IDE isn't VSCode, so I need compilation database to be generated for every project so the LSP would work. The problem is that every time I create/edit project I need to add script which will enable COMPILATIONDB_INCLUDE_TOOLCHAIN. In most cases I don't even need build scripts for anything else besides enabling this flag. I thought maybe it's a good idea to include this flag in `pio run` command? ( e.g. `pio run -t compiledbtc` (?) )

PR #4937, opened 2024-06-16
Add `pio run -t compiledbtc` option

jayadamsmorgan:
Option to generate compile_commands.json with including toolchain paths.

A replacement for having to create pre build script with 'COMPILATIONDB_INCLUDE_TOOLCHAIN=True'

This addresses issue #4769


CodeWenjiu:
This pull request is very useful. Why can't it be merged into the main branch?
Without this, I even get errors when using `std::vector`.

jirutka:
Why is this still not merged? The current recommended approach with `COMPILATIONDB_INCLUDE_TOOLCHAIN` simply doesn’t work.

PR #4947, opened 2024-07-10
Adding Zephyr framework USB VID PID to udev rules.

jalinei:
As Zephyr is officially supported by platformio, it make sense to add its default VID:PID to udev rules.

This will also encompass support for MCUBOOT using zephyr in USB serial mode.


pierre-haessig:
For the reference, this PR is a coordinated effort (in parallel with my documentation edit https://github.com/platformio/platformio-docs/pull/358) to make PlatformIO's doc page about [udev rules](https://docs.platformio.org/en/latest/core/installation/udev-rules.html) a "Single source of truth" to which we can redirect the users of the OwnTech board (which uses Zephyr) from [OwnTech setup page](https://docs.owntech.org/core/docs/environment_setup/). See https://github.com/owntech-foundation/Core/issues/77

ivankravets:
Thanks for the PR!

PR #4952, opened 2024-07-22
allow manual override of system type

shadow578:
adds a check for `PLATFORMIO_SYSTEM_TYPE` environment variable in the system type detection, so that the system type may be defined manually by the user.

this should help in cases where the detection fails, e.g. on windows on arm64 systems that support x86 emulation. 
in those cases, it'd be ok for platformio to pretend like the system is `windows_x86`, since x86 binaries are executed through emulation.

see https://community.platformio.org/t/windows-on-arm64-problem-installing-xtensa-toolchain/25497 for more details.

### DIFF
--- a/platformio/util.py
+++ b/platformio/util.py
@@ -137,11 +137,11 @@ def singleton(cls):
 
 
 def get_systype():
-    # allow manual override, eg. for 
+    # allow manual override, eg. for
     # windows on arm64 systems with emulated x86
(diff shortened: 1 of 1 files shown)


shadow578:
Hi @ivankravets,

is there anything preventing this PR from getting merged? 
judging by the linked issue, it seems like people are beign recommended to patch their platformio installations to include this fix, so it probably works :)

ivankravets:
Thanks for the PR!

PR #4969, opened 2024-08-27
Fix can't match the last+on some boards.

Tokisaki-Galaxy:
### Description

This PR addresses an issue where users are unable to use development boards with '+' at the end of their names. The regular expression used for validating board names has been updated to include the '+' character.

[cannot-matched-issue](https://github.com/platformio/platformio-home/issues/8508)

For example, when I use STC89C516RD+, I will get an error, because the old regular expression regards the+at the end as an error.

### Changes Made
- Updated the regular expression in the configuration file to allow '+' at the end of board names.

### Impact
This change will enable users to use development boards with names ending in '+', which was previously not possible.

### Testing
The board name (STC89C516RD+) with "+" at the end has been tested, and it has been accepted without any problems.

Thank you for considering this PR.

### DIFF
--- a/platformio/project/config.py
+++ b/platformio/project/config.py
@@ -41,7 +41,7 @@ CONFIG_HEADER = """
 
 
 class ProjectConfigBase:
-    ENVNAME_RE = re.compile(r"^[a-z\d\_\-]+$", flags=re.I)
+    ENVNAME_RE = re.compile(r"^[a-z\d\_\-]+(\+)?$", flags=re.I)
     INLINE_COMMENT_RE = re.compile(r"\s+;.*$")
     VARTPL_RE = re.compile(r"\$\{(?:([^\.\}\()]+)\.)?([^\}]+)\}")
 


Tokisaki-Galaxy:
Hi, I noticed an issue that seems to cause significant problems. Users are unable to use development boards with '+' at the end of their names. This can be quite limiting for those who rely on such boards.

I wanted to follow up and ask if there are any concerns or reasons why the PR has not been merged yet.
If there are any changes or improvements needed, please let me know, and I would be happy to make the necessary adjustments.

Thank you for your attention to this matter. I really like this project.

Best regards,
tokisaki

Tokisaki-Galaxy:
Hi, I noticed an issue that seems to cause significant problems. Users are unable to use development boards with '+' at the end of their names. This can be quite limiting for those who rely on such boards.

I wanted to follow up and ask if there are any concerns or reasons why the PR has not been merged yet.
If there are any changes or improvements needed, please let me know, and I would be happy to make the necessary adjustments.

Thank you for your attention to this matter. I really like this project.

Best regards,
tokisaki

ivankravets:
Resolved in https://github.com/platformio/platform-intel_mcs51/commit/ad5d629c010841bbf1ee6f6e65207d8775e4e7cb

calyle:
It never seems to be resolved. Only `STC89C516RD+` can be chose, and there's no `STC89C516RD`.

Tokisaki-Galaxy:
> It never seems to be resolved. Only `STC89C516RD+` can be chose, and there's no `STC89C516RD`.

I believe you are correct. If you look at the commit that was mentioned when this PR was closed, the fix was to rename the board's definition file directly, rather than changing the regular expression as originally proposed.

This is why the STC89C516RD+ board now appears, but the original STC89C516RD is gone. The underlying validation issue still remains.

PR #4974, opened 2024-09-04
add support for `tar.xz` tarballs

Jason2866:
currently `zip` and `tar.gz` as compressed source for packages are supported.
This PR adds support for `tar.xz` compressed tarballs.


ivankravets:
Thanks for the PR!

PR #5043, opened 2024-12-09
Allow HTTP 203 as successful response

maxgerhardt:
Fixes downloading library dependencies from Azure Devops repositories.

See https://community.platformio.org/t/error-downloading-installing-git-repositories-from-behind-a-firewall/9524/2?u=maxgerhardt


ivankravets:
Thanks for the PR!

Issue #5040, opened 2024-12-05
Project Inspector fails when projects are compiled with non-native GCC or non-GCC based compilers


DSann90:
What kind of issue is this?
- [x ] **PlatformIO Core**.
      If you’ve found a bug, please provide an information below.
------------------------------------------------------------------

### Configuration

**Operating system**:
Windows 11 Version 23H2(OS Build 22631)

**PlatformIO Version** (`platformio --version`):
PlatformIO Core, version 6.1.16

### Description of problem
When trying to use the Project Inspector for Memory Analysis with binaries with a non-native GCC or non-GCC based compilers an Assertion Error is thrown and the Memory Analysis is not available.

### Actual Results
COMPILER LOCATION: Compiler set through project option: D:\\uti\\TI\\ARM\\4.0.0.LTS
CONFIGURATION: https://docs.platformio.org/page/boards/timspm0/lp-mspm0g3507.html
PLATFORM: Texas Instruments MSPM0 (0.0.0) > TI Launchpad MSPM0G3507
HARDWARE: 80MHz, 31.25KB RAM, 125KB Flash
DEBUG: Current (openocd_mspm0) On-board (openocd_mspm0) External (jlink)
PACKAGES:
 - openocd_mspm0 @ 1.0.0 
 - tool-jlink @ 1.78811.0 (7.88.11) 
 - toolchain-gccarmnoneeabi @ 1.90301.200702 (9.3.1)

[...]

Generating memory usage report...

*** [sizedata] AssertionError : 
Traceback (most recent call last):
 File \"C:\\Users\\user\\.platformio\\packages\\tool-scons\\scons-local-4.8.1\\SCons\\Action.py\", line 1441, in execute
 result = self.execfunction(target=target, source=rsources, env=env)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File \"C:\\Users\\user\\.platformio\\packages\\tool-scons\\scons-local-4.8.1\\SCons\\Util\\envs.py\", line 252, in __call__
 return self.method(*nargs, **kwargs)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File \"C:\\Users\\user\\.platformio\\penv\\Lib\\site-packages\\platformio\\builder\\tools\\piosize.py\", line 227, in DumpSizeData
 for symbol in _collect_symbols_info(env, elffile, elf_path, sections):
[...]

PR #5051, opened 2024-12-19
Support of non-GCC based compilers for static code analysis (platformio #5040)

DSann90:
Change in the file piosize.py to support nearly all possible compilers for static code analysis using the project inspector.
Commit is suggesting a solution / fix for following Issue: https://github.com/platformio/platformio-core/issues/5040

### DIFF
--- a/platformio/builder/tools/piosize.py
+++ b/platformio/builder/tools/piosize.py
@@ -16,61 +16,82 @@
 
 import json
 import sys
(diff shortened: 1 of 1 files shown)


DSann90:
Hi, can you give me an update about the current state about this pull request if it is fine or when it would probably going to be checked / approved?

PR #5059, opened 2024-12-29
Allow starlette versions through 0.45

musicinmybrain:

https://github.com/encode/starlette/blob/0.45.0/docs/release-notes.md


ivankravets:
Thanks for the PR!

PR #5085, opened 2025-02-13
Clarified language and fixed some grammar issues

Deen-Weible:
I changed some of the markdown (no code) in platformio/project/commands/init.py to be more clear and fix a couple grammar issues.

Noticed them while reading a README it generates and figured I could polish it up!


ivankravets:
Thanks for the PR

PR #5094, opened 2025-02-27
Allow starlette versions through 0.46

musicinmybrain:

https://github.com/encode/starlette/blob/0.46.0/docs/release-notes.md

### DIFF
--- a/platformio/dependencies.py
+++ b/platformio/dependencies.py
@@ -42,7 +42,7 @@ def get_pip_dependencies():
     home = [
         # PIO Home requirements
         "ajsonrpc == 1.2.*",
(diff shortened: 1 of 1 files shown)


ivankravets:
We already bumped a new version in https://github.com/platformio/platformio-core/commit/145142ea6c1d97c92a62999d424c0dd712fbb9f5

You can switch to the development version of PIO Core via `pio upgrade --dev`

ivankravets:
In any case, thanks for the PR and for using PlatformIO!

musicinmybrain:
Thanks! Next time I’ll check the `dev` branch.

PR #5100, opened 2025-03-05
Add support for command length limitation on Windows in CheckToolBase

nachomata:
This pull request includes changes to handle long command lines on Windows by creating a response file when necessary, addressing issue #4446. It is solved by creating a response file when necessary, as suggested [here](https://github.com/platformio/platformio-core/issues/4446#issuecomment-1298958218).

Changes made:

* [`platformio/check/tools/base.py`](diffhunk://#diff-586a52ba8aff19ba1bfefea43cd5947bc1ebd9f15b184c27a02d68cc26949ddfR22): Imported `IS_WINDOWS` from `platformio.compat` to check if the operating system is Windows.
* [`platformio/check/tools/base.py`](diffhunk://#diff-586a52ba8aff19ba1bfefea43cd5947bc1ebd9f15b184c27a02d68cc26949ddfR193-R195): Modified the `execute_check_cmd` method to check if the command length exceeds the Windows limit and create a response file if necessary.
* [`platformio/check/tools/base.py`](diffhunk://#diff-586a52ba8aff19ba1bfefea43cd5947bc1ebd9f15b184c27a02d68cc26949ddfR214-R217): Added the `_create_response_file` helper method to generate a temporary response file containing the command arguments.

This is just a quick fix addressing the immediate problem. It might need some extra testing and fine-tuning, so any suggestions would be welcome.

### DIFF
--- a/platformio/check/tools/base.py
+++ b/platformio/check/tools/base.py
@@ -19,6 +19,7 @@ import click
 
 from platformio import fs, proc
 from platformio.check.defect import DefectItem
+from platformio.compat import IS_WINDOWS
 from platformio.package.manager.core import get_core_package_dir
 from platformio.package.meta import PackageSpec
 from platformio.project.helpers import load_build_metadata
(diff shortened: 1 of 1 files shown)


valeros:
Hi @nachomata, thanks for the PR. I'm afraid none of our backend tools for static code analysis (cppcheck, clang-tidy) supports response files at the moment. For `cppcheck` we already use the `--includes-file` option that helps shorten the final command line. As for `clang-tidy`, there is indeed similar option called `@parameters_file`, but it seems this option cannot be used to pass compilation options (defines, include paths, etc.).

ivankravets:
Thanks for the PR! Duplicate of https://github.com/platformio/platformio-core/issues/5500

Issue #5090, opened 2025-02-16
PIO Core 6.1.17 brakes `compile_commands.json` from `compiledb`


Nathan-ma:
Hi folks, no big deal, reporting an issue with 6.1.17.
After upgrading to 6.1.17, the clangd stopped working as expected; I just downgraded the core version back to `6.1.16` and everything is now working again.

OS: macOS
Basic .ini config:
```
[device]
platform = espressif32@6.4.0
board = esp32dev
framework = arduino
``` 
`.clangd`: 
```
CompileFlags:
  Remove: [-fno-tree-switch-conversion, -mtext-section-literals, -mlongcalls, -fstrict-volatile-bitfields]
```

My LSP when on 6.1.17, fails to find any Arduino framework lib like: WiFi, EEPROM, WiFiProv, etc.

Downgrading to 6.1.16 solves it right away:
```
pip install platformio==6.1.16
```

This is a great opportunity to start supporting version selection with the brew package manager.

ivankravets:
Sorry, I can't reproduce the issue. Could you provide a simple PlatformIO project?

MacMannes:
@ivankravets I have the same problems. I'm using a Mac with NeoVim and the clangd LSP. Here's my project: https://github.com/MacMannes/homey-sonoff-tx

When I use PIO 6.1.18 I get these kind of errors:

<img width="977" alt="Image" src="https://github.com/user-attachments/assets/e2e61996-8f29-4c26-9f3d-8f16a8cfec4e" />

When I downgrade to PIO **6.1.16**, run `pio run -t compiledb` and reopen the `main.cpp` file of my project, the LSP errors disappear. The `compile_commands.json` that are generated between versions **6.1.16** and **6.1.18** are different.

For **6.1.18**, these library paths are missing in the command for `main.cpp`:

-I/Users/andre/.platformio/packages/framework-arduinoespressif8266/libraries/Ethernet/src
-I/Users/andre/.platformio/packages/framework-arduinoespressif8266/libraries/SPI
-I/Users/andre/.platformio/packages/framework-arduinoespressif8266/libraries/ESP8266WiFi/src

proxict:
I believe this issue is related to this: https://github.com/platformio/platformio-core/issues/4998#issuecomment-2848872371

ivankravets:
Please re-test with `pio upgrade --dev`.

Issue #4998, opened 2024-10-14
Space in compilator path not escaped in compilation_db


HanyzPAPU:
- [ ] **PlatformIO Core**.
      If you’ve found a bug, please provide an information below.

------------------------------------------------------------------

### Configuration

**Operating system**:

Windows, with a username who has a space in the username.

**PlatformIO Version** (`platformio --version`):

PlatformIO Core, version 6.1.16

### Description of problem

When the path to the `.platformio` contains a space (for example in the username), the `compile_commands.json` produced by running `pio run -t compiledb` does not escape the compiler executable path in the `command` fields. 

#### Steps to Reproduce

1. Have a user with a space in the username
2. Install PlatformIO, create a new project.
3. Run `pio run -t compiledb`

### Actual Results

Example of a malformed command:

```
[...]

PR #5147, opened 2025-05-04
fix: compiledb doesn't respect build (un)flags (#5090)

proxict:
Related to
 - https://github.com/platformio/platformio-core/issues/5090
 - https://github.com/platformio/platformio-core/issues/4998

This is an attempt to fix the issue of the `compiledb` subcommand not respecting the `build_flags` and `build_unflags` in platformio.ini.

### DIFF
--- a/platformio/builder/tools/piobuild.py
+++ b/platformio/builder/tools/piobuild.py
@@ -58,8 +58,8 @@ def GetBuildType(env):
 
 
 def BuildProgram(env):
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR! 

proxict:
> Please re-test with pio upgrade --dev

```
PlatformIO Core, version 6.1.19a2
```
I can confirm it works as expected now, thanks!

PR #5187, opened 2025-06-23
Add launch configuration and spell check settings for VSCode

cbarney929:


ivankravets:
This is an optional configuration, and you can add it manually.

PR #5204, opened 2025-07-03
Option to disable udev rules check

dimich-dmb:
This introduces "disable_udev_rules_check" option in Linux environment.

If set to True, udev rules are not checked. This allows to suppress annoying warnings when 99-platformio-udev.rules is not installed intentionally.

If set to False (default), udev rules are checked as before.


vidplace7:
Downstream packaging (Fedora, Debian, Arch) would benefit from this change 🙏

ivankravets:
Thanks for the PR!

PR #5226, opened 2025-08-03
Fix: Preserve RFC2217 URLs in SerialPortFinder.find() in "device monitor"

### LINKED ISSUE #5225: rfc2217:// URL dropped by monitor command due to ?
Hello,

When using an RFC2217 remote serial port such as:

```
monitor_port = rfc2217://ras5:4000?ign_set_control
```

PlatformIO’s monitor ignores it and ask for a port:

```
--- Available ports:
--- Enter port index or full name:
```

Even if the port is passed via CLI:

```bash
pio device monitor --port "rfc2217://ras5:4000?ign_set_control"
```

I already found issue in code, I will create a PR next.

Youssefbenhammouda:
Fixes platformio/platformio-core#5225
This PR ensures that fully-qualified serial port URLs (e.g., rfc2217://host:port) are not incorrectly processed as wildcard patterns due to characters like ? or *. Instead, such URLs are returned as-is for serial_for_url() usage.

### DIFF
--- a/platformio/device/finder.py
+++ b/platformio/device/finder.py
@@ -133,6 +133,10 @@ class SerialPortFinder:
 
     def find(self, initial_port=None):
         if initial_port:
+            # Treat any URL (contains '://') as a literal port
+            if "://" in initial_port:
+                return initial_port
+            # Otherwise fall back to existing wildcard logic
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR!

PR #5239, opened 2025-08-29
Fix: exclude nested .vscode directories from package publish

### LINKED ISSUE #5160: Pio pkg publish is including generated files
I noticed while updating my library that pio pkg publish is including the .vscode folder under the project examples, even though these are in the .gitignore and should not be included.

These files contain user names, file names and other generated data and should not be part of the library upload. 

I suggest either:
- checking the .gitignore and respecting the same ignores
- adding an exclude option to the cli
- Providing other means to upload without having these files included

Thank you.

ritesh006:
Fixes #5160

Problem:
`pio pkg pack/publish` included `.vscode` directories inside subfolders
(e.g. examples), even though the top-level `.vscode` was excluded.

Change:
Added `**/.vscode/` (and normalized `.vscode/` with trailing slash) to
EXCLUDE_DEFAULT. This mirrors `.pio` handling and ensures both top-level
and nested `.vscode` folders are excluded automatically.

Manual test:
- Created test package with examples/blink/.vscode/settings.json
- Before patch: file appeared in tarball
- After patch: `.vscode` excluded from tarball

### DIFF
--- a/platformio/package/pack.py
+++ b/platformio/package/pack.py
@@ -49,6 +49,7 @@ class PackagePacker:
         "__*",
         ".DS_Store",
         ".vscode",
+        "**/.vscode/",
         ".cache",
         "**/.cache",
         "**/__pycache__",


ritesh006:
could anyone please review this?

ivankravets:
Thanks for the PR!

PR #5242, opened 2025-08-30
pkg: user-configurable registry mirror (try mirror first, then fallback) (#5235)

### LINKED ISSUE #5235: How to manually specify the download server
I hope to download data from the Github image source because it can be accelerated using Github.

After reading the source code, I found that the variable '__registrant mirror_ hosts__' cannot determine the data download source.

https://dl.registry.platformio.org/download/platformio/tool/toolchain-riscv32-esp/14.2.0+20241119/toolchain-riscv32-esp-linux_x86_64-14.2.0+20241119.tar.gz

But the actual download address is scheduled by the server, which is not good. I hope to manually specify the download source.

ritesh006:
Adds a user-configurable mirror for package downloads.

- New setting: `registry_mirror` (env override: `PLATFORMIO_REGISTRY_MIRROR`)
- _registry.py: prepend user mirror (using artifact filename), then normal mirrors
- _download.py: safe no-op unless filename is provided (prevents bogus paths)
- No behavior change when unset; automatic fallback preserved

Fixes #5235.

### DIFF
--- a/platformio/app.py
+++ b/platformio/app.py
@@ -63,6 +63,15 @@ DEFAULT_SETTINGS = {
         "description": "Verify the proxy server certificate against the list of supplied CAs",
         "value": True,
     },
+    "registry_mirror": {
+        "description": (
+            "Preferred base URL (mirror) for package downloads. "
+            "If set (or overridden by PLATFORMIO_REGISTRY_MIRROR), "
+            "PlatformIO will try this URL first and fall back to the default registry."
+        ),
+        "value": "",
+        "type": "str",
+    },
 }
 
 SESSION_VARS = {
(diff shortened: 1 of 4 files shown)


ritesh006:
Hi.

This PR adds support for a user-configurable registry mirror for package downloads, as discussed in #5235.

Introduces a new registry_mirror setting (with environment override PLATFORMIO_REGISTRY_MIRROR)

When configured, PlatformIO will try the user mirror first (using the artifact filename)

If it fails, the package manager falls back automatically to the default mirrors no behavior change if unset

Minimal, localized changes in app.py, _download.py, base.py, and _registry.py

This gives users in regions with slow or blocked default registry access a way to speed up or stabilize downloads, while preserving existing fallback behavior.

Looking forward to your feedback, happy to refine this if you’d like adjustments

PR #5300, opened 2025-10-28
Update starlette and uvicorn dependencies (upper bounds)

musicinmybrain:
I really care about starlette 0.49 (which contains a security fix, https://github.com/Kludex/starlette/security/advisories/GHSA-7f5h-v6xp-fcq8), but these two dependencies have recently been updated jointly, 6cf8b8172feb33fc2cb59b83f3bb29d44db8922f, so I decided to follow the established pattern.

### DIFF
--- a/platformio/dependencies.py
+++ b/platformio/dependencies.py
@@ -42,8 +42,8 @@ def get_pip_dependencies():
     home = [
         # PIO Home requirements
         "ajsonrpc == 1.2.*",
-        "starlette >=0.19, <0.49",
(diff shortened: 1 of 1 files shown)


musicinmybrain:
Updated to allow [Starlette 0.50.0](https://github.com/Kludex/starlette/releases/tag/0.50.0). The key change is that it drops support for Python 3.9, which has now reached end-of-life.

ivankravets:
Thanks for the PR. Should be updated in the `develop` branch.

PR #5301, opened 2025-10-29
Fix: Handle -U flags with spaces in build_flags

### LINKED ISSUE #5237: Undefines in build_flags don't work if there's a space after -U
I'm on PlatformIO 6.1.18 on Linux, latest version at the time. This is a bug report.

When using -U in build_flags and running a build with --verbose, they're shown in the opening line:

> `Processing WHY2025_badge_companion_radio_ble (extends: WHY2025_badge; build_flags: -w [...] -D LORA_TX_POWER=27, -U LORA_TX_POWER, -D MAX_CONTACTS=100, [...])`

but when gcc is being called, only -U is given to gcc (which incidentally also breaks my build):

`[...] -DARDUINO_PARTITION_default_8MB -U -I.pio/libdeps/WHY2025_badge_companion_rad[...]`

I confirmed this by sneaking in some print debugging to `platformio/builder/tools/piobuild.py`'s `ProcessFlags`:

```python
    ]                                                                                                                                                                                                                
    print("undefines found", undefines)                                                                                                                                                                              
    if undefines:
```

outputs:

```
undefines found ['-U']
```

Removing the space workarounds the issue, but including a space is in the default examples of both gcc and platformio, so I think it's confusing:

- https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html
- https://docs.platformio.org/en/stable/projectconf/sections/env/options/build/build_flags.html

I think this is something that is also a problem for other flags normally, but `ParseFlagsExtended` explicitly fixes it for those classified as `CPPDEFINES` through use of functions like `Util.is_Sequence`.

I think scons.ParseFlags does not explicitly support `-U` (like it does `-D`) and as such is falling back to `CCFLAGS` for it, the corrective code on `ParseFlagsExtended` for `CPPDEFINES` isn't fixing it. Could be escalated to scons, or could be solved in `ParseFlagsExtended` the same way `CPPDEFINES` are fixed.

SpacePlushy:
## Summary

Fixes #5237

When using `-U MACRO` (with space) in `build_flags`, SCons' `ParseFlags()` splits it into two separate items in `CCFLAGS`: `['-U', 'MACRO']`. The previous code only captured `'-U'` and attempted `undef[2:]` which resulted in an empty string.

## Changes

This PR modifies the `ProcessFlags` function in `platformio/builder/tools/piobuild.py` to:

1. **Iterate through CCFLAGS with an index** to detect standalone `-U` flags
2. **Pair `-U` with the following macro name** when they're separate items
3. **Handle both formats**:
   - `-UMACRO` (no space) - already working ✓
   - `-U MACRO` (with space) - now fixed ✓

The solution follows a similar pattern to how SCons handles `-D` flags with spaces.

## Testing

Verified the fix handles:
- Single `-U MACRO` format with space
- Single `-UMACRO` format without space  
- Mixed formats in the same build_flags
- Multiple undefines

## Example

**Before (broken):**
```ini
build_flags = -U LORA_TX_POWER
```
Result: Only `-U` captured, macro name lost

**After (fixed):**
```ini
build_flags = -U LORA_TX_POWER
```
Result: Correctly passes `-U LORA_TX_POWER` to gcc

### DIFF
--- a/platformio/builder/tools/piobuild.py
+++ b/platformio/builder/tools/piobuild.py
@@ -240,17 +240,51 @@ def ProcessFlags(env, flags):  # pylint: disable=too-many-branches
 
     # Cancel any previous definition of name, either built in or
     # provided with a -U option // Issue #191
-    undefines = [
-        u
-        for u in env.get("CCFLAGS", [])
-        if isinstance(u, string_types) and u.startswith("-U")
-    ]
+    ccflags = env.get("CCFLAGS", [])
+    undefines = []
+
+    # Handle both "-UMACRO" and "-U MACRO" formats
+    i = 0
+    while i < len(ccflags):
+        item = ccflags[i]
+        if isinstance(item, string_types) and item.startswith("-U"):
+            if len(item) == 2:  # Just "-U" by itself
+                # Next item should be the macro name
+                if i + 1 < len(ccflags):
+                    macro_name = ccflags[i + 1]
+                    undefines.append(("-U", macro_name))
+                    i += 2  # Skip both "-U" and the macro name
+                    continue
+            else:  # "-UMACRONAME" format
+                macro_name = item[2:]
+                undefines.append(item)
+                i += 1
+                continue
+        i += 1
+
     if undefines:
         for undef in undefines:
-            env["CCFLAGS"].remove(undef)
-            if undef[2:] in env["CPPDEFINES"]:
-                env["CPPDEFINES"].remove(undef[2:])
-        env.Append(_CPPDEFFLAGS=" %s" % " ".join(undefines))
+            # Handle both formats
+            if isinstance(undef, tuple):
+                flag, macro_name = undef
+                env["CCFLAGS"].remove(flag)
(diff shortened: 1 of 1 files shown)


SpacePlushy:
**Note on testing:**

In my initial comment on #5237, I mentioned including tests with this PR. After exploring the codebase, I noticed there aren't existing unit tests for the builder tools (specifically for `ProcessFlags` and related functions in `piobuild.py`).

I didn't want to add a new test structure without understanding your preferred testing approach for the builder module. I'm happy to add tests if you'd like - just let me know:

1. Should I create a new test file for builder tools?
2. Is there a preferred testing pattern you'd like me to follow?
3. Or would you prefer to handle testing separately?

I manually verified the fix handles all edge cases (both `-U MACRO` and `-UMACRO` formats, mixed usage, etc.), but I'm ready to formalize that into proper tests if needed.

ivankravets:
Thanks for the PR. Resolved in https://github.com/platformio/platformio-core/issues/5237

PR #5302, opened 2025-10-29
Fix: Display actual test suite status instead of hardcoded 'PASSED'

### LINKED ISSUE #5183: Unit test runner shows PASSED in summary when there are failing tests
I'm seeing the summary showing as passing when there are failures. Many tests pass, then there's a few failures, and my end-of-test summary looks like this (most names are removed for brevity):
```
...
test/test_math/test_main.cpp:518:PASS
test/test_math/test_main.cpp:519:PASS
test/test_math/test_main.cpp:520:PASS
test/test_math/test_main.cpp:521:PASS
test/test_math/test_main.cpp:522:PASS
test/test_math/test_main.cpp:414:FAIL
test/test_math/test_main.cpp:414:FAIL
test/test_math/test_main.cpp:414:FAIL
test/test_math/test_main.cpp:414:FAIL
test/test_math/test_main.cpp:414:FAIL
test/test_math/test_main.cpp:414:FAIL
test/test_math/test_main.cpp:529:PASS
test/test_math/test_main.cpp:530:PASS
test/test_math/test_main.cpp:531:PASS
...
-----------------------
91 Tests 6 Failures 0 Ignored 
FAIL
```

The final summary shows `PASSED`.

I responded to the following issue, but I don't think this is Teensy-specific: https://github.com/platformio/platform-teensy/issues/68

SpacePlushy:
## Summary

Fixes #5183

The `print_suite_footer` function was hardcoding 'PASSED' for any non-error test suite status, causing SKIPPED test suites to incorrectly display as PASSED in the test footer.

## Changes

Modified `platformio/test/cli.py` to use the actual `test_suite.status.name` with appropriate color coding via `status.to_ansi_color()`.

**Before:**
```python
click.style(test_suite.status.name, fg="red", bold=True)
if is_error
else click.style("PASSED", fg="green", bold=True)  # ← Hardcoded!
```

**After:**
```python
status_color = "red" if is_error else test_suite.status.to_ansi_color()
click.style(test_suite.status.name, fg=status_color, bold=True)
```

## Testing

Verified the fix correctly displays all status types:
- ✅ PASSED (green) - when all tests pass
- ✅ FAILED (red) - when tests fail  
- ✅ ERRORED (red) - when tests error
- ✅ SKIPPED (yellow) - when tests are skipped

### Reproduction
Created test suites with different statuses and verified footer output matches actual test execution results.

**Before fix:** Empty test suite showed `[PASSED]` in footer
**After fix:** Empty test suite correctly shows `[SKIPPED]` in footer

### DIFF
--- a/platformio/builder/tools/piobuild.py
+++ b/platformio/builder/tools/piobuild.py
@@ -240,17 +240,51 @@ def ProcessFlags(env, flags):  # pylint: disable=too-many-branches
 
     # Cancel any previous definition of name, either built in or
     # provided with a -U option // Issue #191
-    undefines = [
-        u
-        for u in env.get("CCFLAGS", [])
-        if isinstance(u, string_types) and u.startswith("-U")
-    ]
+    ccflags = env.get("CCFLAGS", [])
+    undefines = []
+
+    # Handle both "-UMACRO" and "-U MACRO" formats
+    i = 0
+    while i < len(ccflags):
+        item = ccflags[i]
+        if isinstance(item, string_types) and item.startswith("-U"):
+            if len(item) == 2:  # Just "-U" by itself
+                # Next item should be the macro name
+                if i + 1 < len(ccflags):
+                    macro_name = ccflags[i + 1]
+                    undefines.append(("-U", macro_name))
+                    i += 2  # Skip both "-U" and the macro name
+                    continue
+            else:  # "-UMACRONAME" format
+                macro_name = item[2:]
+                undefines.append(item)
+                i += 1
+                continue
+        i += 1
+
     if undefines:
         for undef in undefines:
-            env["CCFLAGS"].remove(undef)
-            if undef[2:] in env["CPPDEFINES"]:
-                env["CPPDEFINES"].remove(undef[2:])
-        env.Append(_CPPDEFFLAGS=" %s" % " ".join(undefines))
+            # Handle both formats
+            if isinstance(undef, tuple):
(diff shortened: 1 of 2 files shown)


ivankravets:
Could you provide a simple project to reproduce this issue?

mahdirajaee (commented):
Good fix. The original code had a clear bug: the else branch hardcoded `click.style("PASSED", fg="green")` regardless of the actual `test_suite.status`, which meant statuses like `SKIPPED`, `WARNED`, or any non-error state all displayed as "PASSED". The new code correctly uses `test_suite.status.name` for the label and delegates to `test_suite.status.to_ansi_color()` for the color, which properly handles all `TestStatus` enum variants.

The one thing I'd verify is the behavior of `to_ansi_color()` for every possible `TestStatus` value — if any status returns `None` or an unexpected value, `click.style(fg=...)` might raise or silently produce uncolored output. A quick check that `TestStatus.SKIPPED.to_ansi_color()` returns a valid click color string would be good to confirm. The `is_error` check is still used correctly for the bar styling via the separate `is_error=is_error` kwarg, so that concern is cleanly separated from the status label. Clean, minimal change that addresses the reported issue.

Note: the diff also includes unrelated changes to `piobuild.py` (handling `-U MACRO` flag formats) that seem to belong to a different PR or concern — might want to split those out to keep the commit focused.

ivankravets:
Thanks for the report! Please re-test with `pio upgrade --dev`.

PR #5326, opened 2025-12-07
Fix a breaking change brought by Python 3.14 asyncio

JakeHarrison11:
Python 3.14 [now returns a `RuntimeError`](https://docs.python.org/3/library/asyncio-eventloop.html#:~:text=Changed%20in%20version%203.14%3A%20Raises%20a%20RuntimeError%20if%20there%20is%20no%20current%20event%20loop.) when `asyncio.get_event_loop()` is called when a loop does not exist. 

This patch creates a loop prior to calling `get_event_loop` so that an error is not thrown and debugging works on versions >3.13.

### DIFF
--- a/platformio/debug/cli.py
+++ b/platformio/debug/cli.py
@@ -167,7 +167,13 @@ def _configure(
 
 
 def _run(project_dir, debug_config, client_extra_args):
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR! Resolved in https://github.com/platformio/platformio-core/commit/41c8e63cd5a1bf0800bb01559b1acd01e97bbc94

PR #5332, opened 2025-12-17
Skip header comments in json manifests

jmwanderer:
Many json files have comments prepended using
the characters # or // comment characters. For example, the json files generated by vscode. A least one manifest parsed by platformio-core, library.json in the Unity, has comments. (See https://github.com/ThrowTheSwitch/Unity/issues/722)

This change skips the header comments in json files.


jmwanderer:
I assume this change should also have a test in tests/package/test_manifiest.py
Perhaps extend test_library_json_parser to include input with comments?

Also, currently this change does not support a file with a comment, white space, and then a comment. Perhaps it will be more robust to add that coverage?

Issue #5334, opened 2025-12-20
Handle comments in JSON library manifests


jmwanderer:
The JSON format does not have a provision for comments, but many users of JSON are prepending comments to JSON files using the // or # comment conventions. (For example, vscode JSON state files are prepended with comments). 

The Unity project has prepended the library.json project with a copyright message in a comment. This breaks the parsing, and will be a problem if PlatformIO upgrades to specifying a new version. Skipping header comments in a JSON file before calling JSON loads seems to be a promising way to fix this.

Specifically:
- scan for comments characters while skipping white space
- if a comment character is found, consume the rest of the line
- stop when a non-comment / non-white space character is found or the end of the string is encountered.

jmwanderer:
Proposed change is in PR #5335

jmwanderer:
I found the comments are removed from library.json during the library install, so this is not actually an issue.

PR #5335, opened 2025-12-20
Skip header comments in json manifests

jmwanderer:
Many json files have comments prepended using
the characters # or // comment characters.

This change skips the header comments in json files.
Updated the Test Manifest to contain a comment line

Proposed fix for https://github.com/platformio/platformio-core/issues/5334


jmwanderer:
I don't believe this is needed at this time. Comments appear to be removed from library.json files at install time.

PR #5336, opened 2025-12-22
Improve UnknownPackageError message clarity

### LINKED ISSUE #4419: Wrong text for UnknownPackageError
https://github.com/platformio/platformio-core/blob/5a0a215bfcee0240018feb6f3ff7f4bce4bcab6a/platformio/package/exception.py#L58

This message is totally not communicative, and it makes think that the architecture is wrong, not that the package is not found.

JoYizPsyKicK:
Fixes #4419

The error message for UnknownPackageError was confusing users by mentioning system architecture, making them think their system was incompatible when the actual issue was simply that the package wasn't found.

Changes:
- Simplified error message to focus on package not found
- Removed unused util import

Testing:
- Code passes pylint with 10/10 score
- Code formatted with black and isort

### DIFF
--- a/platformio/package/exception.py
+++ b/platformio/package/exception.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from platformio import util
 from platformio.exception import UserSideException
 
 
@@ -52,10 +51,7 @@ class MissingPackageManifestError(ManifestException):
 
 
 class UnknownPackageError(PackageException):
-    MESSAGE = (
-        "Could not find the package with '{0}' requirements for your system '%s'"
-        % util.get_systype()
-    )
+    MESSAGE = "Could not find the package with '{0}' requirements"
 
 
 class NotGlobalLibDir(PackageException):
--- a/platformio/package/exception.py
+++ b/platformio/package/exception.py
@@ -54,6 +54,13 @@ class UnknownPackageError(PackageException):
     MESSAGE = "Could not find the package with '{0}' requirements"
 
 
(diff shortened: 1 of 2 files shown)


JoYizPsyKicK:
would like to have someone with the authority to review and act on this pull request

ivankravets:
Thanks for the PR! This PR needs additional work, as UnknownPackageError is generic for both tool-type and library-type packages. While libraries are agnostic to the host machine architecture, “tool” packages are not. We will think further about this issue.

JoYizPsyKicK:
Thank you for the feedback! I understand the concern about UnknownPackageError being too generic for both tool and library packages.
I've updated the implementation to address this:

Changes Made:
Added 
IncompatiblePackageError : A new exception that inherits from UnknownPackageError
Tool-specific error handling : When a tool package is found in the registry but has no compatible version for the user's system architecture, it now raises IncompatiblePackageError with a clear message indicating the system type
Library packages unchanged libraries continue to use UnknownPackageError as they are architecture-agnostic

JoYizPsyKicK:
This approach maintains backward compatibility (since IncompatiblePackageError inherits from UnknownPackageError) while providing clearer feedback to users when architecture incompatibility is the issue.

Please let me know if this addresses your concerns or if you'd like me to take a different approach!

JoYizPsyKicK:
@ivankravets 
Following up on this , would appreciate a review from someone who can take a decision on this PR.

ivankravets:
Thanks for the PR! Please re-test with `pio upgrade --dev`

JoYizPsyKicK:
Thank you for merging the PR!

PR #5339, opened 2025-12-29
twisted.python.constants is moving to constantly, try both

markmentovai:
References:

https://docs.twisted.org/en/twisted-16.5.0/core/howto/constants.html
https://github.com/twisted/twisted/issues/12361
https://github.com/twisted/twisted/pull/12362
https://github.com/twisted/twisted/commit/6bf282ed1aee9135ac4c76139d01b012024e3104 (2024-11-12), in https://github.com/twisted/twisted/releases/tag/twisted-24.11.0 (2024-12-02)


ivankravets:
Thanks for the PR!

PR #5341, opened 2026-01-01
Update 99-platformio-udev.rules

VinzRep:
Using an one-more Arduino Uno chinese clone, it use FX2348N chip for USB to Serial communication.

I'm new in linux and open-source contributor, and I spend some times to understand what's wrong.

This is my first pull request in my life :)


ivankravets:
Thanks for the PR!

PR #5350, opened 2026-01-11
Allow Starlette 0.51, 0.52

musicinmybrain:
In Starlette [0.51](https://github.com/Kludex/starlette/blob/0.51.0/docs/release-notes.md#0510-january-10-2026), the  warning stacklevel on `DeprecationWarning` for the wsgi module was increased. There should be no impact to platformio-core.


ivankravets:
Thanks for the PR. Please re-test with `pio upgrade --dev`

PR #5360, opened 2026-01-14
Remove static code analysis via PVS-Studio

SigmaTel71:
Relates to https://github.com/platformio/platformio-docs/pull/396

***

Removes static analysis via PVS-Studio support as PVS-Studio had ended support for PlatformIO:
> **Note:** Integration with PlatformIO was available in earlier versions but is no longer officially supported.

They had updated their [old integration overview article](https://pvs-studio.com/en/blog/posts/cpp/0714/) to reflect these changes.


SigmaTel71:
Rebased against the latest `develop` commit. There was a merge conflict I didn't have time to handle. 😅

ivankravets:
Thanks for the PR!

PR #5407, opened 2026-03-26
Scope `build_src_flags` to project source compilation only

### LINKED ISSUE #5245: Bug: build_src_flags has an impact on lib_deps
`$ pio --version`
PlatformIO Core, version 6.1.19a2

see my https://github.com/platformio/platformio-core/issues/4667#issuecomment-3240460103

in particular this is the case for "-Wpedantic" set in `build_src_flags`

huyhoang171106:
## Summary

The current build flow applies `build_src_flags` too early (on a shared/base construction environment), which leaks those flags into library compilation (`lib_deps`). This breaks third-party libraries when users set strict flags like `-Wpedantic` in `build_src_flags`. The fix is to keep `build_src_flags` isolated and apply them only when compiling files from `$PROJECT_SRC_DIR` (and not to library builders, framework packages, or other dependency compilation units).

## Files changed

- `platformio/builder/tools/piobuild.py` (modified)

## Testing

Closes #5245

### DIFF
--- a/platformio/builder/tools/piobuild.py
+++ b/platformio/builder/tools/piobuild.py
@@ -176,7 +176,9 @@ def ProcessProjectDeps(env):
             env.Exit(1)
 
     if "test" not in env["BUILD_TYPE"] or env.GetProjectOption("test_build_src"):
-        plb.env.BuildSources(
+        src_env = plb.env.Clone()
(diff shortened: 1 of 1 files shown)


ivankravets:
Spam

PR #5408, opened 2026-03-26
Expand conditional LDF scanner include paths to use full compile-time include context

### LINKED ISSUE #4818: LDF doesn't resolve dependencies in (+) modes if an intermediate header file is located outside lib folders
### Description of problem

LDF (in any mode that evaluates C/C++ Preprocessor conditional syntax) is not able to find a dependency when a header from that dependency is hidden under an `ifdef` guard and the macro that enables the include header file is located outside search paths (framework libraries, `include` and `src` folders) passed to `CCONDITIONAL_SCANNER`,

For example the `CONFIG_ETH_ENABLED` macro in the code below is set in the `sdkconfig.h` file which is located in the framework folder `$PIO_PACKAGES/framework-arduinoespressif32/tools/sdk/esp32/dio_qspi/include`.

#### Steps to Reproduce

Compile the following code:

```cpp
#include <Arduino.h>
#include <sdkconfig.h>

#ifdef CONFIG_ETH_ENABLED
#include "ETH.h"
#endif

void setup() {}

void loop() {}
```

with the following configuration:
```ini
[env:esp32dev]
platform = espressif32
framework = arduino
board = esp32dev
lib_ldf_mode = deep+
```

### Actual Results

```
Scanning dependencies...
No dependencies
Compiling .pio\build\esp32dev\src\main.cpp.o
src/main.cpp:25:10: fatal error: ETH.h: No such file or directory
```
### Expected Results

```
Scanning dependencies...
Dependency Graph
|-- Ethernet @ 2.0.0
```

huyhoang171106:
## Summary

In `deep+`/`chain+`-style LDF modes, dependency resolution currently fails when a macro-gating header (for example `sdkconfig.h`) is not inside the limited scanner roots (`src`, `include`, framework-lib folders). The conditional scanner evaluates `#ifdef` blocks without seeing macro definitions from that external header, so guarded includes (like `ETH.h`) are skipped and the dependency graph is incomplete.

## Files changed

- `platformio/builder/tools/piolib.py` (modified)

## Testing

Closes #4818

### DIFF
--- a/platformio/builder/tools/piolib.py
+++ b/platformio/builder/tools/piolib.py
@@ -120,6 +120,16 @@ class LibBuilderBase:
 
     _INCLUDE_DIRS_CACHE = None
 
+    def get_conditional_scanner_cpppath(self):
+        cpppath = []
+        for item in self.envorigin.get("CPPPATH", []) + self.env.get("CPPPATH", []):
+            if item not in cpppath:
+                cpppath.append(item)
+        for item in self.get_include_dirs():
+            if item not in cpppath:
+                cpppath.append(item)
+        return cpppath
+
     def __init__(self, env, path, manifest=None, verbose=False):
         self.env = env.Clone()
         self.envorigin = env.Clone()


ivankravets:
> get_conditional_scanner_cpppath() is never called, 

spam?

PR #5411, opened 2026-03-26
Introduce dynamic IDE project-file basename derived from project directory

### LINKED ISSUE #4264: control names of IDE project files
What kind of issue is this?

- [x] **Feature Request**.

### Description of problem

At least for Qt Creator, the project name follows the name of the generated `*.creator` file (and associated files) rather than the directory name.  So if `pio project init` always uses the basename `platformio` for generated files, all the projects will have the same name in the IDE.

Since there are now multiple files generated for qtcreator, renaming is tedious (though I can do it with a utility like `rename`).

### Expected Results

Perhaps `project init` should name the IDE project files following the directory name.

huyhoang171106:
## Summary

The current integration flow appears to rely on a fixed basename (`platformio`) when producing IDE artifacts. Add a single source of truth in the integration generator for the IDE filename stem, defaulting to the current project directory name (not a hardcoded constant). This enables all IDE integrations that use this basename (including Qt Creator) to generate uniquely named files per project.

## Files changed

- `platformio/project/integration/generator.py` (modified)

## Testing

Closes #4264

### DIFF
--- a/platformio/project/integration/generator.py
+++ b/platformio/project/integration/generator.py
@@ -64,6 +64,10 @@ class ProjectGenerator:
             ]
         )
 
+    def get_project_files_basename(self):
+        basename = os.path.basename(os.path.abspath(self.project_dir))
+        return basename or os.path.basename(os.getcwd())
+
     @staticmethod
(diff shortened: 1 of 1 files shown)


ivankravets:
AI spam

PR #5412, opened 2026-03-26
fix integrity.dat not being cleaned on version string typo

### LINKED ISSUE #5232: integrity.dat can't be cleansed if typo happened on versioning string of platformio.ini
So, if I do the typo on versioning string of platformio.ini, in my case I hit 'm' key instead of ','  , it will be "carved" into the integrity.dat in the .pio>libdeps>(envname)>integrity.dat, and can't be normally cleaned off with clean build.
since I typo-ed when editing platformio.ini, I had to manually delete integrity.dat for it to regenerate to solve the polluted versioning string which was stopping the build.
it would be nice if integrity.dat was regularly regenerated to fit the content of platform.ini? I'm still beginner so not really sure how to fix it but it's a potensional trap for someone. I was a bit confused because ctrl + shift + s search on vscode don't show the content inside '.' prefixed folders.  
P.S.
Also this is the first issues I create in my life so please be patient if some formatting was wrong or something. thanks!

knQzx:
catch SemanticVersionError when removing unused libdeps so that a typo in the version string doesn't permanently corrupt integrity.dat and block builds

fixes #5232

### DIFF
--- a/platformio/package/commands/install.py
+++ b/platformio/package/commands/install.py
@@ -20,6 +20,7 @@ import click
 
 from platformio import fs
 from platformio.package.exception import UnknownPackageError
+from platformio.package.version import SemanticVersionError
 from platformio.package.manager.core import get_core_package_dir
 from platformio.package.manager.library import LibraryPackageManager
 from platformio.package.manager.platform import PlatformPackageManager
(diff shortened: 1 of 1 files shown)


ivankravets:
Thanks for the PR!

PR #5423, opened 2026-03-31
fix: usage with starlette >= 1

grawlinson:
In 1.0.0, starlette has [removed the on_startup/shutdown handlers](https://starlette.dev/release-notes/#100rc1-february-23-2026) (Kludex/starlette#3117) in favour of a [lifespan handler](https://www.starlette.dev/lifespan/). This commit implements support for starlette >= 1.

However, I have not touched `platformio/dependencies.py` since I do not know what this project's policy is on dependencies. Let me know if you'd like that implemented also.

### DIFF
--- a/platformio/home/run.py
+++ b/platformio/home/run.py
@@ -14,6 +14,8 @@
 
 import os
 from urllib.parse import urlparse
+import contextlib
+import contextvars
 
 import click
 import uvicorn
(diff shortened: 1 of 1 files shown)


platformio/home/run.py
```diff
@@ -86,16 +87,22 @@ def run_server(host, port, no_open, shutdown_timeout, home_url):
     if path != "/":
         routes.append(Route("/", protected_page))
 
+    @contextlib.asynccontextmanager
+    async def lifespan(app):
+        # startup
+        click.echo("PIO Home has been started. Press Ctrl+C to shutdown.")
+
+        if(not no_open):
```
dekeonus: neither `no_open` nor `home_url` are available to lifespan() here.

dekeonus:
I added support for starlette-1.0.0 to gentoo's build for platformio-core.
see:
https://github.com/gentoo/gentoo/blob/28f6b4b79aad17eb8938bd70cd3fa1911dd51469/dev-embedded/platformio/files/pio-6.1.19-starlette-1.patch

platformio/home/run.py
grawlinson: Fixed with recent rebase.

grawlinson:
> I added support for starlette-1.0.0 to gentoo's build for platformio-core. see: https://github.com/gentoo/gentoo/blob/28f6b4b79aad17eb8938bd70cd3fa1911dd51469/dev-embedded/platformio/files/pio-6.1.19-starlette-1.patch

I've just replaced the PR with this commit, as it's a lot better ... not to mention that it actually works!

ivankravets:
Thanks for the PR. Please re-test with `pio upgrade --dev`

PR #5435, opened 2026-04-18
Avoid infinite loop when cached mirror redirect keeps failing

bdraco:
The registry mirror iterator caches the HEAD redirect for one hour, but the cache hit path never advanced `_visited_mirrors`, and the cached payload did not include `X-PIO-Mirror`. When a prior run cached a redirect to a mirror that now fails (for example an SSL error on the download), subsequent iterations kept hitting the same cache key and returning the same failing mirror forever, so tool installs looped until interrupted.

The fix stores `X-PIO-Mirror` in the cached payload, and on cache hit only reuses the entry if it names a mirror not already tried; otherwise it falls through to a fresh HEAD with `bypass` set. Stale cache entries written before this change lack `X-PIO-Mirror` and fall through naturally, so no manual cache clear is required.

### Before

```
Tool Manager: Installing platformio/tool-esptoolpy @ ~1.30000.0
Tool Manager: Error: Please read https://bit.ly/package-manager-ioerror
Tool Manager: Warning! Package Mirror: HTTPSConnectionPool(host='usc1.contabostorage.com', port=443): Max retries exceeded with url: /.../tool-esptoolpy-1.30000.201119.tar.gz (Caused by SSLError(SSLCertVerificationError(...)))
Tool Manager: Looking for another mirror...
Tool Manager: Error: Please read https://bit.ly/package-manager-ioerror
Tool Manager: Warning! Package Mirror: HTTPSConnectionPool(host='usc1.contabostorage.com', port=443): Max retries exceeded with url: /.../tool-esptoolpy-1.30000.201119.tar.gz (Caused by SSLError(SSLCertVerificationError(...)))
Tool Manager: Looking for another mirror...
... (repeats indefinitely against the same host until the user interrupts)
```

### After

```
Tool Manager: Installing platformio/tool-esptoolpy @ ~1.30000.0
Tool Manager: Error: Please read https://bit.ly/package-manager-ioerror
Tool Manager: Warning! Package Mirror: HTTPSConnectionPool(host='usc1.contabostorage.com', port=443): Max retries exceeded with url: /.../tool-esptoolpy-1.30000.201119.tar.gz (Caused by SSLError(SSLCertVerificationError(...)))
Tool Manager: Looking for another mirror...
Unpacking  [####################################]  100%
Tool Manager: tool-esptoolpy@1.30000.201119 has been installed!
```

The iterator now advances to a working mirror after the first failure, and if every mirror fails the loop terminates with a clear error instead of spinning forever.

### DIFF
--- a/platformio/registry/mirror.py
+++ b/platformio/registry/mirror.py
@@ -43,10 +43,17 @@ class RegistryFileMirrorIterator:
             if result is not None:
                 try:
                     headers = json.loads(result)
-                    return (
-                        headers["Location"],
-                        headers["X-PIO-Content-SHA256"],
-                    )
+                    mirror = headers.get("X-PIO-Mirror")
+                    # Only use the cached entry if it identifies a mirror we
+                    # have not already tried. Without this, a cached redirect
+                    # to a failing mirror would be returned forever because
+                    # _visited_mirrors never advances on the cache-hit path.
+                    if mirror and mirror not in self._visited_mirrors:
+                        self._visited_mirrors.append(mirror)
+                        return (
+                            headers["Location"],
+                            headers["X-PIO-Content-SHA256"],
+                        )
                 except (ValueError, KeyError):
                     pass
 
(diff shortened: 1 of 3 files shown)


bdraco:
This has been a night and day reliability improvement. No more having to delete the cache every few hours

ivankravets:
Hi Nick,

Thanks for the PR! 🙏  I'll review it.

ivankravets:
Thanks for the PR!

bdraco:
Thanks!

PR #5448, opened 2026-05-22
app: use explicit ValueError instead of assert in projects_dir_validate (fixes #5403)

### LINKED ISSUE #5403: fix: validation relies on assert, which is stripped in optimized python mode
## Description

projects_dir_validate() uses `assert os.path.isdir(projects_dir)`. In production environments running Python with `-O`, asserts are removed, so invalid paths are silently accepted. That bypasses config validation and causes downstream failures in unrelated code paths, making errors harder to diagnose.

**Severity**: `medium`
**File**: `platformio/app.py`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

adityachilka1:
Fixes #5403.

### The bug

`platformio/app.py:`

```python
def projects_dir_validate(projects_dir):
    assert os.path.isdir(projects_dir)
    return os.path.abspath(projects_dir)
```

Python drops `assert` statements under `-O` (and when the `PYTHONOPTIMIZE`
environment variable is set). On production images and distro packages that run
Python in optimised mode, any string - even a fake path - gets through the
validator untouched. The error then surfaces downstream (e.g. when the projects
directory is later globbed), where it is harder to diagnose as a config issue.

### The fix

Replace the `assert` with an explicit `raise ValueError`. One-line diff:

```diff
 def projects_dir_validate(projects_dir):
-    assert os.path.isdir(projects_dir)
+    if not os.path.isdir(projects_dir):
+        raise ValueError("Not a directory: %s" % projects_dir)
     return os.path.abspath(projects_dir)
```

### Why ValueError

Caller (`sanitize_setting`, platformio/app.py:168) wraps any exception from
validators into `exception.InvalidSettingValue` via `raise ... from exc`. So
the user-facing error path is identical. `ValueError` is idiomatic for
"argument has the correct type but wrong value".

### Tests

Adds `tests/misc/test_app.py` with three cases:

#### Verification (local)

```
python -O -c "from platformio import app; app.projects_dir_validate('/nope')"
# Traceback (most recent call last):
# ...
# ValueError: Not a directory: /nope
```

Also verified that the pre-fix code (replicated in a sandbox) silently returned
an invalid path under `-O`.

### No behavioural regression

- Normal (non-optimised) runs: previously `assert` raised `AssertionError`
  (a subclass of `Exception`) which `sanitize_setting` caught. Now raises
  `ValueError` which it still catches. No user-facing difference.
- Optimised (`-O`) runs: was a silent pass-through. Now correctly raises.

### Related

The CONTRIBUTING.md requires a CLA. I'm filing this as an individual contributor
(@adityachilka1) and will sign the CLA through CLA-assistant when the bot prompts.

### DIFF
--- a/platformio/app.py
+++ b/platformio/app.py
@@ -29,7 +29,8 @@ from platformio.project.helpers import get_default_projects_dir
 
 
 def projects_dir_validate(projects_dir):
-    assert os.path.isdir(projects_dir)
+    if not os.path.isdir(projects_dir):
+        raise ValueError("Not a directory: %s" % projects_dir)
     return os.path.abspath(projects_dir)
 
 
--- a/tests/misc/test_app.py
+++ b/tests/misc/test_app.py
@@ -1,0 +1,48 @@
+# Copyright (c) 2014-present PlatformIO <contact@platformio.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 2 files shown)


ivankravets:
Duplicate of https://github.com/platformio/platformio-core/issues/5403

Issue #5305, opened 2025-11-04
Does not work correctly in pipx/uvx-style environments due to missing pip dependency


cgranade:
When using `uvx` or `uv tool install` to run PlatformIO, many commands fail due to a missing dependency on `pip`:

```
$ mkdir esp-example
$ cd esp-example
$ uvx platformio init --board esp32s3usbotg
The following files/directories have been created in /home/cgranade/tmp/esp-example
include - Put project header files here
lib - Put project specific (private) libraries here
src - Put project source files here
platformio.ini - Project Configuration File
Resolving esp32s3usbotg dependencies...
Platform Manager: Installing espressif32
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
Platform Manager: espressif32@6.12.0 has been installed!
Tool Manager: Installing espressif/toolchain-xtensa-esp32s3 @ 8.4.0+2021r2-patch5
Downloading  [####################################]  100%          
Unpacking  [####################################]  100%          
Tool Manager: toolchain-xtensa-esp32s3@8.4.0+2021r2-patch5 has been installed!
Tool Manager: Installing espressif/toolchain-riscv32-esp @ 8.4.0+2021r2-patch5
Downloading  [####################################]  100%          
Unpacking  [####################################]  100%          
Tool Manager: toolchain-riscv32-esp@8.4.0+2021r2-patch5 has been installed!
Tool Manager: Installing platformio/framework-arduinoespressif32 @ ~3.20017.0
Downloading  [####################################]  100%          
Unpacking  [####################################]  100%          
Tool Manager: framework-arduinoespressif32@3.20017.241212+sha.dcc1105b has been installed!
Tool Manager: Installing platformio/tool-esptoolpy @ ~2.40900.0
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
/home/cgranade/.cache/uv/archive-v0/UqYPTk0VrTbywRTRcX1sL/bin/python: No module named pip


Please ensure that the following packages are installed:

sudo apt install python3-dev libffi-dev libssl-dev


Cleaning...
Tool Manager: tool-esptoolpy@2.40900.250804 has been installed!
MissingPackageManifestError: Could not find one of 'package.json' manifest files in the package
```

The problem can be confirmed by noting that re-running `uvx` with the pip dependency added manually causes the command to complete successfully:

```
$ uvx --with pip platformio init --board esp32s3usbotg
Installed 24 packages in 37ms
[...]

PR #5468, opened 2026-07-05
Declare pip runtime dependency

### LINKED ISSUE #5418: PlatformIO fails in pipx/uvx environments — pip not declared as a dependency
## Description

PlatformIO's tool manager internally calls `pip install` to install Python-based tool dependencies (e.g., esptool crypto dependencies), but `pip` is not listed in `platformio/dependencies.py`.

Modern Python tool runners (`pipx`, `uvx`, `uv tool install`) create isolated virtual environments that only include declared dependencies. Since `pip` is not declared, it is absent from the environment, and tool installation fails with:

```
ModuleNotFoundError: No module named 'pip'
```

## Reproduction

```bash
pipx install platformio
pio run -e esp32dev  # fails when trying to install esptool dependencies
```

## Proposed Fix

**Option A (one-line fix):** Add `pip` to `get_pip_dependencies()` in `dependencies.py`:

```python
def get_pip_dependencies():
    return [
        "pip",
        # ... existing dependencies
    ]
```

**Option B (long-term):** Refactor the tool manager to use `importlib` and direct wheel installation instead of shelling out to `pip`. This removes the runtime dependency entirely.

## Context

Related to #5305. The Python ecosystem is rapidly moving toward `uv`-based tooling, making this an increasingly common failure mode for new users.

JamBalaya56562:
## Summary

- declare `pip` as a PlatformIO Python dependency so pipx/uvx/uv tool environments install it explicitly
- remove the duplicate hard-coded `pip` upgrade target from `upgrade_pip_dependencies()`
- add focused tests for the dependency list and upgrade command construction

## Root cause

PlatformIO invokes `python -m pip` at runtime for tool/Python package installation, but `pip` was not listed in `get_pip_dependencies()`. Isolated runners such as pipx, uvx, and `uv tool install` only install declared dependencies, so those environments can omit the `pip` module and fail with `No module named pip`.

Fixes #5418.
Related to #5305.

## Validation

- `C:\Users\Jam\AppData\Local\mise\installs\uv\0.11.26\uv.exe run --with pytest --with-editable . -- python -m pytest tests/misc/test_dependencies.py`
- `C:\Users\Jam\AppData\Local\mise\installs\uv\0.11.26\uv.exe run --with-editable . -- python -c "import importlib.metadata as md; reqs=md.requires('platformio') or []; print([r for r in reqs if r == 'pip']); print(len(reqs))"`
- `mise exec uv@0.11.26 -- uv run --with pytest --with-editable . -- python -m pytest tests/misc/test_dependencies.py`

### DIFF
--- a/platformio/commands/upgrade.py
+++ b/platformio/commands/upgrade.py
@@ -111,7 +111,6 @@ def upgrade_pip_dependencies(verbose):
             "pip",
             "install",
             "--upgrade",
-            "pip",
             *get_pip_dependencies(),
         ],
         check=True,
--- a/platformio/dependencies.py
+++ b/platformio/dependencies.py
@@ -28,6 +28,7 @@ def get_core_dependencies():
 
 def get_pip_dependencies():
     core = [
+        "pip",
         "bottle == 0.13.*",
         "click >=8.0.4, <8.5, !=8.3",  # click 9.0 removes 'protected_args' attribute
         "colorama",
--- a/tests/misc/test_dependencies.py
+++ b/tests/misc/test_dependencies.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2014-present PlatformIO <contact@platformio.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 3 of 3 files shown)


ivankravets:
Thanks for the PR! Please re-test with `pio upgrade --dev`.

PR #5474, opened 2026-07-10
Preserve user .vscode/extensions.json on IDE regeneration (#5473)

### LINKED ISSUE #5473: project init --ide vscode unconditionally overwrites .vscode/extensions.json, destroying user ordering, formatting, and comments
## Description

When regenerating VS Code integration files, PlatformIO **unconditionally rewrites** `.vscode/extensions.json` from a template on every index rebuild. It does not edit the existing file in place — it re-renders the whole file from scratch and overwrites whatever was there.

Because the template is the sole source of truth for the file's structure, every rebuild reverts anything the user customized:

- **Ordering** — output is always `sorted()`, so any intentional ordering is undone.
- **Formatting** — canonical 4-space indentation, fixed header, fixed key layout are re-imposed.
- **Comments** — the user's comments are stripped and never re-emitted (only the template's own two header comment lines are written back).

The template *does* attempt to preserve the *set* of recommendation names by reading and merging the existing file, so in steady state on a PlatformIO-generated file the resulting bytes may be identical — but the write still happens unconditionally, so the file's mtime changes and git/editors flag it as modified on every rebuild. And the moment the user customizes anything about the file, that customization is reverted on the next rebuild.

There is a related, narrower bug in the merge step: the comment-stripping only handles full-line `//` comments, so **inline `//` or `/* */` comments cause the merge to fail silently and the user's recommendation *entries* to be deleted entirely** (details below). But this is secondary — even a perfect parser wouldn't fix the primary problem, because the file is re-rendered from a template rather than edited in place.

## Expected behavior

- **If `.vscode/extensions.json` already exists:** PlatformIO should perform an **in-place insertion** — ensure `platformio.platformio-ide` is present in `recommendations` (and `ms-vscode.cpptools-extension-pack` in `unwantedRecommendations`), and otherwise leave the file completely untouched: preserve the user's ordering, formatting, **and comments**. If the entries are already present, the file should not be written at all.
- **If the file does not exist:** scaffolding a default from the template (as today) is fine.

Preserving comments requires the file to be treated as **JSONC** (VS Code's actual format for `extensions.json`), so comment-aware parsing is a necessary part of the fix, not just a nice-to-have.

## Affected code

**Writer:** `platformio/project/integration/generator.py`, `_merge_contents()` — always opens in `"w"` mode and writes, with no comparison to existing content. Note there's already a skip guard for `.gitignore`, but nothing equivalent for other files:

```python
@staticmethod
def _merge_contents(dst_path, contents):
    if os.path.basename(dst_path) == ".gitignore" and os.path.isfile(dst_path):
        return
    with open(dst_path, "w", encoding="utf8") as fp:   # always writes
        fp.write(contents)
```

**Template:** `platformio/project/integration/tpls/vscode/.vscode/extensions.json.tpl` — re-renders the entire file from a `set` of recommendations rendered via `sorted()`, and re-emits its own fixed header comments. The merge step strips comments with a regex that only handles full-line `//`:

```python
% recommendations = set(["platformio.platformio-ide"])
% unwantedRecommendations = set(["ms-vscode.cpptools-extension-pack"])
% previous_json = os.path.join(project_dir, ".vscode", "extensions.json")
% if os.path.isfile(previous_json):
%   fp = open(previous_json)
%   contents = re.sub(r"^\s*//.*$", "", fp.read(), flags=re.M).strip()
%   fp.close()
%   if contents:
%       try:
%           data = json.loads(contents)
%           recommendations |= set(data.get("recommendations", []))
%           unwantedRecommendations |= set(data.get("unwantedRecommendations", []))
%       except ValueError:
%           pass
```

`re.sub(r"^\s*//.*$", ...)` matches only whole-line `//` comments. Inline `//` and `/* */` comments survive, break `json.loads`, and the `except ValueError: pass` silently swallows the failure — collapsing `recommendations` back to just the template default.

## Steps to reproduce

**Primary problem (overwrite of ordering/formatting/comments):**

1. Create a project and run `pio project init --ide vscode`.
2. Edit `.vscode/extensions.json` — reorder entries, adjust formatting, and add a comment:
   ```jsonc
   {
       // Recommended for our team
       "recommendations": [
           "esbenp.prettier-vscode",
           "platformio.platformio-ide"
       ]
   }
   ```
3. Run `pio project init --ide vscode` again (or trigger an index rebuild from the VS Code extension).

**Expected:** the file is left as-is (all required entries already present).
**Actual:** the file is rewritten — comment removed, entries re-sorted, formatting normalized to the template.

**Secondary problem (silent entry deletion via comments):**

Add an inline or block comment alongside a custom recommendation, then rebuild:
```jsonc
{
    "recommendations": [
        "platformio.platformio-ide",
        "esbenp.prettier-vscode" // team formatter
    ]
}
```
**Expected:** `esbenp.prettier-vscode` is preserved.
**Actual:** `recommendations` is reset to just `["platformio.platformio-ide"]`; the user entry is gone.

## Minimal proof (secondary bug)

Isolating the template's merge logic (against `develop` @ `a0efc7f4`, v6.2.0a3):

```
Case 1 (clean, extra rec):      ['esbenp.prettier-vscode', 'platformio.platformio-ide']
Case 2 (full-line // comment):  ['esbenp.prettier-vscode', 'platformio.platformio-ide']
Case 3 (inline // comment):     ['platformio.platformio-ide']   <-- user rec LOST
Case 4 (/* block */ comment):   ['platformio.platformio-ide']   <-- user rec LOST
```

## Impact

- **Loss of user customization:** ordering, formatting, and comments in a user-curated (and often VCS-committed) `.vscode/extensions.json` are reverted on every rebuild.
- **Data loss:** with inline/block comments present, custom recommendation entries are silently deleted — no error, no warning.
- **VCS churn:** the file is rewritten on every rebuild even when semantically unchanged, producing noisy diffs on shared `.vscode/extensions.json`.

## Suggested direction

The goal is only to *ensure* `platformio.platformio-ide` (and the unwanted-recommendation default) are present — a membership insertion, not a full-file regeneration. Proposed fix (happy to submit a PR):

1. **When the file exists, insert in place.** Parse the existing `extensions.json`, and only if a required entry is missing, add just that array element while preserving the user's comments, ordering, and formatting. If nothing needs adding, don't write the file at all.
2. **Parse as JSONC.** Handle inline `//` and `/* */` comments (comment-tolerant load / edit) so comments are preserved and the merge never silently drops entries.
3. **When the file does not exist, scaffold from the template** as today.

## Environment

- PlatformIO Core: 6.2.0a3 (`develop`, commit `a0efc7f4`)
- Reproducible independent of OS (pure template/parsing logic)

tillig:
## Description

Fixes #5473.

Regenerating the VSCode integration (`pio project init --ide vscode`, which the VSCode extension triggers on every IntelliSense index rebuild) **unconditionally overwrote** an existing `.vscode/extensions.json` from a template. This had two consequences:

1. **Loss of user customization / VCS churn.** The file was re-rendered from the template on every rebuild, reverting the user's ordering, formatting, and comments, and rewriting the file even when nothing had semantically changed.
2. **Silent data loss.** The template merged the existing file by stripping comments with `re.sub(r"^\s*//.*$", ...)` (full-line `//` only) and then `json.loads`. An **inline `//`** or **`/* */` block comment** (both valid JSONC, which VSCode supports for this file) left invalid JSON, `json.loads` raised `ValueError`, and `except ValueError: pass` swallowed it — collapsing `recommendations` back to just the PlatformIO default and **deleting the user's own recommendations**.

## Changes

- **New helper `platformio/project/integration/jsonc.py`** — minimal JSONC support: comment/trailing-comma-tolerant `loads()`, and `ensure_array_items()` which inserts missing array entries **in place**, preserving the surrounding comments, ordering, and formatting.
- **`ProjectGenerator._merge_contents`** — for `extensions.json`, when the file already exists the required entries are inserted in place; the file is only rewritten when an entry is actually added. If the existing file cannot be parsed as JSONC, it safely falls back to regenerating from the template. When the file is absent, it is scaffolded from the template as before. Other generated files (`c_cpp_properties.json`, `launch.json`) are unaffected and still regenerate.
- **Template** simplified to a static scaffold (the parse/merge logic now lives in code).
- **Tests** — unit tests for the JSONC helper (comments, block comments, trailing commas, comment markers inside strings, empty arrays, missing keys, no-op when already present, invalid input) and end-to-end `project init --ide vscode` tests covering (a) an existing customized file left byte-for-byte unchanged on re-run and (b) a missing recommendation inserted while the user's inline comment and existing entry survive.
- **HISTORY.rst** entry under 6.2.0.

## Why a hand-rolled JSONC helper instead of a dependency?

The problem is not *parsing* JSONC — it's **editing it in place without destroying the user's comments**. Standard JSONC parsers (`json5`, `pyjson5`, `commentjson`, `hjson`) only parse to a Python object; comments do not survive into the parsed result, so a load → mutate → dump round-trip would reintroduce the exact data loss this PR fixes. Comment-preserving round-trip editing of JSONC has no established Python library (unlike `ruamel.yaml` for YAML). Since a parser dependency wouldn't remove the need for the in-place insertion logic — and PlatformIO keeps its dependency set intentionally small — the helper parses only to *decide* whether an entry is missing, then performs a surgical string insertion into the original text, leaving every other byte untouched. It is pure-Python, has no I/O, and is covered by unit tests. Happy to swap in a maintainer-preferred approach if you'd rather.

## Checklist

- [x] `make before-commit` (codespell, isort, black, pylint) passes on the changed files (pylint 10.00/10)
- [x] Tests added and passing
- [ ] CLA signed

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -34,6 +34,7 @@ test-driven methodologies, and modern toolchains for unrivaled success.
 * Fixed an issue where the test suite status was incorrectly hardcoded as PASSED (`issue #5183 <https://github.com/platformio/platformio-core/issues/5183>`_)
 * Fixed an issue where using a relative path with the ``-d, --project-dir`` option caused `pio device monitor <https://docs.platformio.org/en/latest/core/userguide/device/cmd_monitor.html>`__ to fail with a "FileNotFoundError" (`pull #5470 <https://github.com/platformio/platformio-core/pull/5470>`_)
 * Fixed compatibility with ``pipx``- and ``uvx``-managed environments by ensuring the required ``pip`` dependency is available when installing Python-based tool packages (`issue #5305 <https://github.com/platformio/platformio-core/issues/5305>`_)
+* Fixed an issue where regenerating the VSCode integration overwrote an existing ``.vscode/extensions.json``, discarding the user's comments, ordering, and formatting, and silently dropping recommendations when the file contained inline or block comments; the required entries are now inserted in place (`issue #5473 <https://github.com/platformio/platformio-core/issues/5473>`_)
 
 6.1.19 (2026-02-04)
 ~~~~~~~~~~~~~~~~~~~
--- a/platformio/project/integration/generator.py
+++ b/platformio/project/integration/generator.py
@@ -21,6 +21,7 @@ from platformio import fs, util
 from platformio.debug.helpers import get_default_debug_env
 from platformio.proc import where_is_program
 from platformio.project.helpers import load_build_metadata
+from platformio.project.integration import jsonc
 
 
 class ProjectGenerator:
@@ -169,9 +170,48 @@ class ProjectGenerator:
         with open(tpl_path, "r", encoding="utf8") as fp:
             return bottle.template(fp.read(), **tpl_vars)
 
-    @staticmethod
-    def _merge_contents(dst_path, contents):
-        if os.path.basename(dst_path) == ".gitignore" and os.path.isfile(dst_path):
+    # File name -> {JSONC key: [required values]}. For these files an existing
+    # user file is updated in place (missing entries are inserted) instead of
+    # being regenerated, so the user's comments, ordering, and formatting are
+    # preserved.
+    IN_PLACE_MERGE_RULES = {
+        "extensions.json": {
+            "recommendations": ["platformio.platformio-ide"],
+            "unwantedRecommendations": ["ms-vscode.cpptools-extension-pack"],
+        }
+    }
+
+    @classmethod
+    def _merge_contents(cls, dst_path, contents):
+        file_name = os.path.basename(dst_path)
+        if file_name == ".gitignore" and os.path.isfile(dst_path):
             return
+        merge_rules = cls.IN_PLACE_MERGE_RULES.get(file_name)
+        if merge_rules and os.path.isfile(dst_path):
+            if cls._merge_jsonc_in_place(dst_path, merge_rules):
+                return
         with open(dst_path, "w", encoding="utf8") as fp:
             fp.write(contents)
+
+    @staticmethod
+    def _merge_jsonc_in_place(dst_path, merge_rules):
+        """Ensure the required entries exist in an existing JSONC file without
+        touching the user's comments, ordering, or formatting. Only rewrites
+        the file when something was actually inserted. Returns ``True`` when the
+        existing file was handled (and must not be regenerated), ``False`` when
+        it could not be parsed and should fall back to regeneration."""
+        with open(dst_path, "r", encoding="utf8") as fp:
+            original = fp.read()
+        try:
+            updated = original
+            changed = False
+            for key, values in merge_rules.items():
+                updated, key_changed = jsonc.ensure_array_items(updated, key, values)
+                changed = changed or key_changed
+        except ValueError:
+            # Not parseable as JSONC; let the caller regenerate from template.
+            return False
+        if changed:
+            with open(dst_path, "w", encoding="utf8") as fp:
+                fp.write(updated)
+        return True
--- a/platformio/project/integration/jsonc.py
+++ b/platformio/project/integration/jsonc.py
@@ -1,0 +1,183 @@
+# Copyright (c) 2014-present PlatformIO <contact@platformio.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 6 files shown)


ivankravets:
Resolved in https://github.com/platformio/platformio-core/commit/e23eafb963ad3043d32a41bdaf3ec2e7dd02c19e

PR #5480, opened 2026-07-23
builder: include test sources in compiledb via __test target

### LINKED ISSUE #4934: `pio run -t compiledb` does not generate compilation database for tests
What kind of issue is this?

- [ ] **Question**.
      This issue tracker is not the place for questions. If you want to ask how to do something,
      or to understand why something isn't working the way you expect it to, 
      use [Community Forums](https://community.platformio.org) or [Premium Support](https://platformio.org/support)

- [ ] **PlatformIO IDE**.
      All issues related to PlatformIO IDE should be reported to the
      [PlatformIO IDE for VSCode](https://github.com/platformio/platformio-vscode-ide/issues) repository

- [ ] **Development Platform or Board**.
      All issues (building, uploading, adding new boards, etc.) related to PlatformIO development platforms
      should be reported to appropriate repository related to your hardware
      https://github.com/topics/platformio-platform

- [ ] **Feature Request**.

- [x] **PlatformIO Core**.

*You can erase any parts of this template not applicable to your Issue.*

### Configuration

**Operating system**: Arch Linux

**PlatformIO Version** (`platformio --version`): PlatformIO Core, version 6.1.14

### Description of problem
I use pio run -t compiledb to generate a compilation database that clangd can utilise. However, I noticed that the generated database does not include the path to unity.h and hence I get linter errors in my editor (neovim) whenever I try to include unity.h.

image

#### Steps to Reproduce

1. `pio run -t compiledb`

### Actual Results
Compilation database only generated for project root.

### Expected Results
Compilation database should also be generated for tests.

Socialpranker:
### AI disclosure

Written by Claude (Anthropic, Sonnet 5). The entire patch, root-cause analysis, and verification below were produced by an AI agent. I (the human account owner) reviewed the diff and the verification steps before submitting.

### Problem

`pio run -t compiledb` leaves test sources out of `compile_commands.json`, so clangd/editors can't resolve `unity.h` (or any test-framework include) inside `test/` files. Adding `-t __test` to opt into the test build doesn't help either — it fails with `Nothing to build. Please put your test suites to the 'test' folder` even when suites exist in the documented `test/test_*/` layout.

### Root cause (two layers)

1. **`ConfigureTestTarget` default filter isn't recursive.** With no `PIOTEST_RUNNING_NAME` set (it's only passed by `pio test`, per suite), `PIOTEST_SRC_FILTER` defaults to `+<*.c>`-style patterns, which `fs.match_src_files` treats as *root-level only* (a glob that matches a directory recurses; one that matches files doesn't). The documented layout puts sources in `test_*/` subdirectories, so the filter matches nothing → "Nothing to build".

2. **`__test` aliases to the full build.** Even with sources found, `builder/main.py` aliases `__test` to `DEFAULT_TARGETS`, so `pio run -t compiledb -t __test` would attempt to compile and *link* every suite — each with its own `main()` — into one program and fail at the link step. Compiledb generation itself never compiles or links anything, so this build isn't needed.

### Fix

- In `ConfigureTestTarget`, when no suite is selected **and** `compiledb` is among the command-line targets, include every `test_*/` suite recursively in `PIOTEST_SRC_FILTER` and add each suite dir to `CPPPATH` (mirroring what `pio test` does per suite).
- In `builder/main.py`, when `compiledb` is requested, alias `__test` to nothing instead of the default build targets.

Both changes are scoped to the compiledb code path: a plain `pio run -t compiledb`, a plain `pio run -t __test`, and `pio test` all keep their exact current behavior.

### Usage after this fix

```
pio run -t compiledb -t __test
```

produces a `compile_commands.json` containing the test suites plus the test framework sources (e.g. `Unity/src/unity.c`, whose flags carry the `unity.h` include path the issue reporter needed). With `test_build_src = yes` in `platformio.ini`, production sources are included too.

### Verification

All checks below were run for real on macOS with `platform = native` (no hardware needed):

**Red** (unpatched develop): `pio run -t compiledb -t __test` on a project with `test/test_calc/test_add.c` → `Nothing to build` / FAILED. Plain `pio run -t compiledb` → only `src/` files in the database.

**Green** (this patch): same command → SUCCESS, `compile_commands.json` contains `test/test_main/test_add.c`, `Unity/src/unity.c`, and the generated `unity_config.c`; with `test_build_src = yes`, `src/` files are present as well.

**Non-regression**:
- Plain `pio run -t compiledb` → unchanged (only `src/`, no test sources) — covered by a dedicated test.
- `pio test -e native` on the same project → PASSED, identical before and after the patch.
- Full `tests/commands/test_test.py` suite (9 tests, including the `examples/unit-testing/calculator` end-to-end test) → all pass with the patch.

Added `tests/commands/test_run_compiledb.py` with 3 tests (test sources included; test+src with `test_build_src = yes`; plain compiledb unchanged). Verified they fail on unpatched develop (2 failed / 1 passed — the non-regression one) and pass with the patch. `black`, `isort`, and `pylint --rcfile=./.pylintrc` are clean on all three changed files (10.00/10).

Fixes #4934

### DIFF
--- a/platformio/builder/main.py
+++ b/platformio/builder/main.py
@@ -205,7 +205,15 @@ env.AddPreAction(
 )
 
 AlwaysBuild(env.Alias("__debug", DEFAULT_TARGETS))
-AlwaysBuild(env.Alias("__test", DEFAULT_TARGETS))
+if "compiledb" in COMMAND_LINE_TARGETS:
+    # `pio run -t compiledb -t __test` (issue #4934): "__test" is passed only
+    # to include test sources in the compilation database. Don't alias it to
+    # the default build targets — a full build would try to link every test
+    # suite (each with its own `main()`) into one program and fail, while
+    # compiledb generation itself never compiles or links anything.
+    env.Alias("__test", [])
+else:
+    AlwaysBuild(env.Alias("__test", DEFAULT_TARGETS))
 
 env.ProcessDelayedActions()
 
--- a/platformio/builder/tools/piotest.py
+++ b/platformio/builder/tools/piotest.py
@@ -14,6 +14,8 @@
 
 import os
 
+from SCons.Script import COMMAND_LINE_TARGETS  # pylint: disable=import-error
+
 from platformio.builder.tools import piobuild
 from platformio.test.result import TestSuite
 from platformio.test.runners.factory import TestRunnerFactory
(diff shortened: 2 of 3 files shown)


Socialpranker:
Friendly ping — open since 23 July, CLA signed and no conflicts.

It gets test sources into `compile_commands.json` so clangd can resolve `unity.h` and friends inside `test/`. Today `pio run -t compiledb` skips them, and opting in with `-t __test` fails with `Nothing to build. Please put your test suites to the 'test' folder` even for the documented `test/test_*/` layout.

Happy to split the two layers (the `ConfigureTestTarget` guard and the compiledb wiring) into separate PRs if that's easier to review.

Socialpranker:
Bumping this — about two weeks now, CLA's signed and it's still mergeable as far as I can tell.

PR #5485, opened 2026-07-30
Modernize deprecated ::set-output to \

KarloffsGhost:
GitHub deprecated the ::set-output workflow command in favor of writing to the $GITHUB_OUTPUT environment file. GitHub has since postponed actual removal indefinitely, so this workflow still runs fine today with only a deprecation warning — this is a forward-compatible modernization, not a repair of something currently broken.

This is a mechanically verified, byte-precise, line-for-line replacement of both ::set-output call sites in this file with the equivalent $GITHUB_OUTPUT form. The transform was checked for syntax correctness, shell-context safety, runner-capability compatibility, and re-scanned afterward to confirm zero remaining deprecated call sites, then confirmed the resulting YAML still parses. No other content in the file was touched.

Full evidence, embedded here so nothing depends on an external link:

- Repository commit verified: `a0efc7f48566c967333d505086973354b60a35a0`
- File: `.github/workflows/docs.yml`
- SHA-256 before patch: `2939df4d0d6db14ee3ac402cc46ef9689511e38ba6519d02a7a5c5a9270519c5`
- SHA-256 after patch: `a3b20f6a291294769858f10343b7c2dd05c78b0f6846b943623e6c597f764da2`
- Exact diff:
```diff
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -72,9 +72,9 @@ jobs:
         id: get-destination-dir
         run: |
           if [[ ${{ env.RELEASE_BUILD }} == true ]]; then
-            echo "::set-output name=dst_dir::stable"
+            echo "dst_dir=stable" >> "$GITHUB_OUTPUT"
           else
-            echo "::set-output name=dst_dir::latest"
+            echo "dst_dir=latest" >> "$GITHUB_OUTPUT"
           fi
       - name: Checkout latest Docs
         continue-on-error: true
```

Happy to close this if it's not wanted or if there's a preferred way to handle the migration.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -72,9 +72,9 @@ jobs:
         id: get-destination-dir
         run: |
           if [[ ${{ env.RELEASE_BUILD }} == true ]]; then
-            echo "::set-output name=dst_dir::stable"
+            echo "dst_dir=stable" >> "$GITHUB_OUTPUT"
           else
-            echo "::set-output name=dst_dir::latest"
+            echo "dst_dir=latest" >> "$GITHUB_OUTPUT"
           fi
       - name: Checkout latest Docs
         continue-on-error: true


ivankravets:
Thanks for the PR!

PR #5494, opened 2026-08-06
Sort project library integrity data

### LINKED ISSUE #5491: integrity.dat under libdeps seems to be order dependent?
I have been investigating an issue with our docker images whereby running 

```
platformio pkg install 
platformio run
```
will cause the library dependencies to be reinstalled. From what I found, they can produce different orders in the resulting `integrity.dat` file.  

In the platformio.ini, I have:
```
libdeps = 
    https://github.com/thomasfredericks/Bounce2#d744e0690f5b5d403f2847014fa1b16c72da9970
    solarcaratuva/boost-preprocessor@^1.77.0 
```

`platformio pkg install`  creates an integrity.dat file with the same order:
```
https://github.com/thomasfredericks/Bounce2#d744e0690f5b5d403f2847014fa1b16c72da9970
solarcaratuva/boost-preprocessor@^1.77.0 
```

Whereas `platformio run` creates an integrity.dat file with a different ordering:
```
solarcaratuva/boost-preprocessor@^1.77.0
https://github.com/thomasfredericks/Bounce2#d744e0690f5b5d403f2847014fa1b16c72da9970
```

Changing the order of my `libdeps` to match the resolution order of `platformio run` fixes the problem for `pkg install` but I think they should could just sort the `integrity.dat` before checking for matches.

## Version
PlatformIO Core, version 6.1.19

Bortlesboat:
Closes #5491

Serialize the unique project library specifications in sorted order when
writing `integrity.dat`. This keeps identical dependency sets byte-stable across
Python processes without changing dependency resolution, installation order,
set comparison, or unused-package removal.

The focused regression exercises the cleanup helper and asserts the canonical
file contents; it fails against the previous implementation under a hash seed
that produces the opposite set order.

Validation:

- `tests/commands/pkg/test_install.py`: 11 passed
- `tox run -e lint`: 10.00/10 for source and tests
- `git diff --check`

AI assistance: OpenAI Codex helped draft the change and regression test. I
reviewed the complete diff and validation evidence and take responsibility for
the contribution.

### DIFF
--- a/platformio/package/commands/install.py
+++ b/platformio/package/commands/install.py
@@ -287,7 +287,7 @@ def _uninstall_project_unused_libdeps(lm, lib_deps):
                 pass
     if not storage_dir.is_dir():
         storage_dir.mkdir(parents=True)
-    integrity_dat.write_text("\n".join(lib_deps), encoding="utf-8")
+    integrity_dat.write_text("\n".join(sorted(lib_deps)), encoding="utf-8")
 
 
 def _install_project_private_library_deps(private_pkg, private_lm, env_lm, options):
--- a/tests/commands/pkg/test_install.py
+++ b/tests/commands/pkg/test_install.py
@@ -20,7 +20,10 @@ import pytest
 
 from platformio import fs
 from platformio.dependencies import get_core_dependencies
-from platformio.package.commands.install import package_install_cmd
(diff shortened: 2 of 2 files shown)


ivankravets:
7*8=?

PR #5499, opened 2026-08-12
prevent path traversal via chmod and mtime in zip unpacking

knQzx:
`ZIPArchiver` applied chmod and mtime to the raw archive entry name joined onto the destination directory, so an entry like `../name` could change permissions (including setuid) on a file outside the extraction directory. the tar path already guards this with `is_bad_path`, the zip path did not

this lifts `is_bad_path`/`resolve_path` to the base archiver and blocks zip entries that resolve outside the destination, matching the existing tar behavior


ivankravets:
Thanks for the PR!

PR #5507, opened 2026-08-30
Update marshmallow constraints (allow 3.x and 4.x)

vidplace7:
Marshmallow 3.x is not being packaged anymore in latest Fedora/Arch/others.
This has caused the Fedora `python-platformio` package to become broken / orphaned.

Update constraints to allow Marshmallow 3.x and 4.x.

https://marshmallow.readthedocs.io/en/stable/upgrading.html#upgrading-to-4-0


ivankravets:
Thanks for the PR!

PR #5514, opened 2026-09-14
ci: split PLATFORMIO_CI_SRC on os.pathsep

r3wretrhy:
`pio ci` reads `PLATFORMIO_CI_SRC` and splits it on `:`. On Windows the native path separator is `;`, and a drive letter is already a colon, so `C:\proj\src\main.cpp` becomes `['C', '\proj\src\main.cpp']` and `validate_path` rejects it.

Split on `os.pathsep` instead. POSIX still uses `:`; Windows keeps the drive letter and splits multiple roots on `;`.

The added unit tests fail with a literal `:` split and pass with `os.pathsep`.


ivankravets:
Thanks for the PR. Left the code simple. Please re-test with `pio upgrade --dev`.

PR #5517, opened 2026-09-18
Fail loudly on invalid local dependency paths

### LINKED ISSUE #4901: Silent failure when dependency path is incorrect in library.json
### Configuration

**Operating system**: Windows

**PlatformIO Version**: 6.1.14

### Description of problem

If you put this in a `library.json` file:

```
{
  "dependencies": {
    "nonexistent_lib_1": "symlink:../hfzehfzeiohize",
    "nonexistent_lib_2": "file:../hrfiohgeuibz"
  }
}
```

PlatformIO will happily tell you:
```
Library Manager: Installing symlink:../hfzehfzeiohize
Library Manager: Installing file:../hrfiohgeuibz
```
even though none of those paths exist. This makes tracking path issues painful.

### Expected Results

I would expect:
```
Library Manager: Installing symlink:../hfzehfzeiohize
PackageException: Can not create a symbolic link for `../hfzehfzeiohize`, not a directory
```
or:
```
Library Manager: Installing symlink:../hfzehfzeiohize
VCSBaseException: VCS: Unknown repository type symlink:../hfzehfzeiohize
```

Mawasi1:
Resolve #4901

`symlink:../lib` / `file:../lib` in library.json `dependencies` was parsed by `PackageSpec._parse_owner` as an owner/name pair, so it became a registry lookup. That fails with `UnknownPackageError`, which `install_dependencies` swallows for deps without an `owner` key. Result: `pio pkg install` silently skipped the dependency, while `pio run` hard-failed for the same library under the project `lib/` folder.

Changes:
- `_parse_owner`: don't split as owner/name when the text before the slash contains a colon. The value now reaches `install_from_uri` and is rejected with the existing `VCS: Unknown repository type` error, consistently in both commands.
- `install_from_uri`: raise a descriptive `PackageException` for a missing `file://` path instead of a raw `FileNotFoundError` traceback, matching the `symlink://` branch.
- Tests for both, offline.

Regular specs (`owner/name`, `owner/name@ver`, `name=owner/pkg`, git/zip URLs, `file:///`, `symlink://`, bare names) parse identically before and after.

Behavioural notes, both intentional:
- `pio run` now fails with the same error when any installed library (project `lib/` or `.pio/libdeps`) declares a malformed local-path dependency. Previously the `lib/` case hard-failed with `UnknownPackageError` and the `libdeps` case built silently without the dependency.
- Other strings with a colon before the single slash (e.g. `C:/lib`, `host:8080/pkg`) were already failing as registry lookups for a non-existent owner; they now fail with the URI error instead. Existing Windows paths are unaffected because `_parse_local_file` converts them to `file://` first.

Supersedes #5024, which changed `project/options.py` rather than the package manager.

AI disclosure: written with Claude Code assistance; I reproduced the issue, reviewed the diagnosis and the diff, and ran the tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ivankravets:
I can only imagine what will happen to the OSS projects if the maintainers start merging every piece of nonsense that AI generates.

PR #5518, opened 2026-09-18
Fixed `-U` undefine flag being dropped when separated from its name b…

### LINKED ISSUE #5237: Undefines in build_flags don't work if there's a space after -U

Isl-n:
…y a space // Resolve #5237


ivankravets:
How did you find PlatformIO?

ivankravets:
Thanks for the PR. Resolved in https://github.com/platformio/platformio-core/issues/5237

Isl-n:
I found PlatformIO while working on an ESP32 project at a course and really liked the ecosystem. Thanks for building it—I'll keep an eye out for other open issues to contribute to!

ivankravets:
I like the solution in this PR. Did you use any AI model?

Isl-n:
Yes, Here is my exact workflow:
1- Used AI code review and search tools to scan the repository for open bugs/features to work on.
2- Selected this issue and worked through the fix using Claude Sonnet.
3- Ran pytest locally to verify the solution and make sure all tests passed green.
4- Submitted the PR once everything was fully tested!

